PR #82, opened 2020-11-25
feat: enable multiple AWS rules on WAF

AntoineAugusti:
Use a common set of rules from AWS on the WAF 

These rules are also used by [the COVID Alert server](https://github.com/cds-snc/covid-alert-server-staging-terraform/blob/master/server/aws/waf.tf)

### DIFF
--- a/aws/eks/waf.tf
+++ b/aws/eks/waf.tf
@@ -6,8 +6,10 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
     allow {}
   }
 
+  # Use a bunch of AWS managed rules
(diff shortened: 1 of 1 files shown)


dorianjp (approved):
im assuming as these are pre-defined rule-sets, the config is close to boilerplate from aws?

thanks!

AntoineAugusti:
@dorianjp Yeah exactly. Did you look at the doc https://docs.aws.amazon.com/waf/latest/developerguide/aws-managed-rule-groups-list.html? There is a link in the code. They have a list of managed rules and I picked what seems to be useful for us.

PR #86, opened 2020-11-26
feat: add alarms for SMS success rate

AntoineAugusti:
Here is what the graph looks fine over the last months in production
Screen Shot 2020-11-26 at 13 31 42


brdunfield (approved):
LGTM 👍 

Do we know if AWS has any thresholds / cutoffs for SMS delivery like they do for email?

AntoineAugusti:
@brdunfield I'm not aware of anything like this and I didn't see it in the documentation. I'll open a support case to have a clear answer 👌 

If they have one, we'll adjust this

brdunfield:
Sounds good. These thresholds look fine based on our past data, and we've never to my knowledge received any complaints from AWS about it, so I am not too worried, just curious.

PR #90, opened 2020-11-30
feat: switch to 24 hours for SMS success rates

AntoineAugusti:
Adjust the period for alarms related to success rates for SMS as we currently have low volumes


brdunfield (approved):
👍 Iterate all the things!

PR #91, opened 2020-12-01
feat: add CI script to ask to update alarms sheet

AntoineAugusti:
This PR adds a CI script to ask us to update the Google Sheet when requesting a review when editing an alarms' file.

I tested it by doing a dummy change, requesting a review and then reverting the change


brdunfield (approved):
👍 This is a great idea!
🦑

Issue #94, opened 2020-12-01
Hook up ALB logs to S3


maxneuvians:
Currently the ALB is not logging into S3. We have a bucket, but no logs https://s3.console.aws.amazon.com/s3/buckets/notification-canada-ca-production-alb-logs?region=ca-central-1&tab=objects

PR #95, opened 2020-12-01
feat: output ALB log bucket

AntoineAugusti:
Towards #94

In order to set up [access logs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb#access_logs) for the load balancer, we need to output the name of the bucket to reuse it in the other module (S3 bucket in the `common` module while the load balancer is in the `eks` module).

I've adjusted the lifecycle rule to delete logs after 90 days - we keep cluster logs forever, I'm not sure we need that much to troubleshoot.


brdunfield (approved):
🦑

PR #97, opened 2020-12-02
feat: log ALB requests to S3

### LINKED ISSUE #94: Hook up ALB logs to S3

AntoineAugusti:
Fixes #94 and uses the output introduced in #95


brdunfield (approved):
🦑

PR #99, opened 2020-12-03
fix: add bucket policy for S3 ALB bucket

AntoineAugusti:
Turns out #97 is not enough, the [CI job failed](https://github.com/cds-snc/notification-terraform/actions/runs/398860370). So still working on #94 

After [reading the doc](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html#access-logging-bucket-permissions) one more time, we need to grant permissions to some AWS-owned accounts to allow load balancers to write objects in our bucket


dorianjp (approved):
Thanks for the view.

Good to see the way a bucket policy works.

PR #100, opened 2020-12-03
Bump production to 0.44.1

AntoineAugusti:
Release the log bucket for the load balancer to production #94


AntoineAugusti:
I skipped it because we tagged the two very closely!

Issue #96, opened 2020-12-02
Add a log metric to count/isolate services over daily rate limit


AntoineAugusti:
Log messages https://github.com/cds-snc/notification-api/blob/1fd2b427b8cf878bf92c54c04a22ddb3f231cf1b/app/notifications/validators.py#L36-L56

Log metrics code to get some inspiration
https://github.com/cds-snc/notification-terraform/blob/09a706c646cdf3163c7d266ad74b1f03dd32982e/aws/eks/cloudwatch_log.tf#L13-L23

After the log metric has been added, maybe add a graph on a dashboard and see if we can get log messages based on this

PR #101, opened 2020-12-08
feat: track number of services going over the rate limit

jimleroyer:
### Issue

https://trello.com/c/62gscLJH/169-graph-the-services-going-over-the-daily-rate-limit

### Purpose

As a business owner of the Notify platform.
I want to track services going over the daily rate limit,
so that I can monitor usages and contact the teams overusing the service.

As a sysops responsible for the Notify platform,
I want to track services going over the daily rate limit,
so that I make sure the SLAs are respected and individual services do not hinder usage of overall platform (especially for other unrelated services).

### Changes 

Adds a metric to AWS CloudWatch for us to graph in our Notify accounts. 

### Tests

Creating this branch and see how it runs should tell us if that works out.

### DIFF
--- a/aws/eks/cloudwatch_log.tf
+++ b/aws/eks/cloudwatch_log.tf
@@ -33,3 +33,15 @@ resource "aws_cloudwatch_log_metric_filter" "celery-error" {
     value     = "1"
   }
 }
+
+resource "aws_cloudwatch_log_metric_filter" "over-rate-limit" {
+  name           = "over-rate-limit"
+  pattern        = "has been rate limited for daily use sent"
+  log_group_name = "/aws/containerinsights/${aws_eks_cluster.notification-canada-ca-eks-cluster.name}/application"
+
+  metric_transformation {
+    name      = "over-rate-limit"
+    namespace = "LogMetrics"
+    value     = "1"
+  }
+}
--- a/aws/eks/cloudwatch_log.tf
+++ b/aws/eks/cloudwatch_log.tf
@@ -34,8 +34,8 @@ resource "aws_cloudwatch_log_metric_filter" "celery-error" {
   }
 }
 
-resource "aws_cloudwatch_log_metric_filter" "over-rate-limit" {
-  name           = "over-rate-limit"
+resource "aws_cloudwatch_log_metric_filter" "over-daily-rate-limit" {
+  name           = "over-daily-rate-limit"
   pattern        = "has been rate limited for daily use sent"
   log_group_name = "/aws/containerinsights/${aws_eks_cluster.notification-canada-ca-eks-cluster.name}/application"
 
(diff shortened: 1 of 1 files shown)


aws/eks/cloudwatch_log.tf
```diff
@@ -33,3 +33,15 @@ resource "aws_cloudwatch_log_metric_filter" "celery-error" {
     value     = "1"
   }
 }
+
+resource "aws_cloudwatch_log_metric_filter" "over-rate-limit" {
+  name           = "over-rate-limit"
```
AntoineAugusti, line 38: Could you rename to `over-daily-rate-limit`

There is also a rate limit at the API level, if you do too many calls quickly

aws/eks/cloudwatch_log.tf
```diff
@@ -33,3 +33,15 @@ resource "aws_cloudwatch_log_metric_filter" "celery-error" {
...
+  pattern        = "has been rate limited for daily use sent"
```
AntoineAugusti, line 39: Can you try to test if this works without adding double quotes to the pattern? I see that other metric filters have double-quotes.

You can try with another log message or look in the documentation to confirm it works fine

AntoineAugusti (commented):
Good work! 👏 

Minor things before approving

AntoineAugusti:
Fixes #96

aws/eks/cloudwatch_log.tf
jimleroyer: Alright done.

aws/eks/cloudwatch_log.tf
jimleroyer: The AWS documentation says these are not necessary if it's only alpha-numeric characters or underscore:
> Metric filter terms that include characters other than alphanumeric or underscore must be placed inside double quotes ("").

I'd say it's probably a best practice to keep these as the absence of double quotes don't seem to offer advantages if I get this right, whereas not having these could lead to unexpected results.

AntoineAugusti:
@jimleroyer You forgot to push your commits I think

aws/eks/cloudwatch_log.tf
```diff
@@ -33,3 +33,15 @@ resource "aws_cloudwatch_log_metric_filter" "celery-error" {
...
   }
 }
+
+resource "aws_cloudwatch_log_metric_filter" "over-daily-rate-limit" {
+  name           = "over-daily-rate-limit"
+  pattern        = "has been rate limited for daily use sent"
+  log_group_name = "/aws/containerinsights/${aws_eks_cluster.notification-canada-ca-eks-cluster.name}/application"
+
+  metric_transformation {
+    name      = "over-rate-limit"
```
AntoineAugusti, line 43:
```suggestion
    name      = "over-daily-rate-limit"
```

AntoineAugusti (approved):
Accept my suggestion and then good to go!

Thanks and congrats on your first PR here 👏

PR #105, opened 2020-12-10
Bump production to 0.46

AntoineAugusti:
Compare URL https://github.com/cds-snc/notification-terraform/compare/v0.44.1...v0.46.0

This will release to production: a new metric filter already tested in staging by @jimleroyer and a bunch of firewall rules on the load balancer. No requests will be blocked yet - blocked that would have been blocked will only be counted for the moment, we'll review rules if necessary (as we did in #107)


dorianjp (approved):
Looks good.

PR #107, opened 2020-12-14
feat: enable requests blocking with WAF

AntoineAugusti:
Rules added in https://github.com/cds-snc/notification-terraform/pull/82 were only counting blocked requests

Start blocking requests matching one of these rules


aws/eks/waf.tf
```diff
@@ -3,7 +3,7 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
   scope = "REGIONAL"
 
   default_action {
-    allow {}
+    block {}
```
dorianjp, line 6: trying to understand the scope of the default action, so this will enable the blocking to occur when matched against the rules below?

AntoineAugusti: You're right. I may have this wrong. I find the documentation really difficult to understand https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl

I may need to set the default action to `allow` and add an action set to block for every rule.

Do you understand the same thing?

AntoineAugusti:
@dorianjp After reading the documentation one more time I switched to what I described in this message https://github.com/cds-snc/notification-terraform/pull/107#discussion_r542646607 and I think this is the way to go.

It's important to point out that this change will go in staging first and only unless we explicitly bump the production tag.

Issue #117, opened 2020-12-21
Add a redirection from www subdomain


AntoineAugusti:
http://www.notification.canada.ca does not resolve or redirects at the moment. Some people think that all URLs must start with `www` and will land on an error page.

Make sure we have DNS records for this domain and redirect to the non www version at the load balancer level.

This will require a PR to [the DNS repository](https://github.com/cds-snc/dns) as well.

PR #118, opened 2020-12-21
feat: redirect www domain to non www

AntoineAugusti:
Fixes https://github.com/cds-snc/notification-terraform/issues/117

`priority` on the load balancer is set this way currently:
- values < 100 are for redirect rules
- values > 100 are used to dispatch requests to target groups

DNS PR: https://github.com/cds-snc/dns/pull/171


jimleroyer (approved):
LGTM

PR #122, opened 2020-12-23
feat: add missing warning alarms

AntoineAugusti:
3 alarms only had critical alarms and not an associated warning one. We want to make sure that we don't discover important issues too late, that's why it's important to have a warning for each thing we take a close look at. Warning alarms have lower thresholds and will trigger before any critical alarms. We use these alarms to investigate blips before they become issues.

To determine the missing warning alarms, I looked at the code and the sheet.


jimleroyer (approved):
Super!

PR #126, opened 2021-01-05
feat: set up SNS SMS in us-west-2

AntoineAugusti:
Set up SNS text messages in us-west-2, to be able to send text messages with long code numbers ordered through Pinpoint in the us-west-2 region.

This PR:
- set SMS preferences, introducing a new variable for spending limit in the us-west-2 region
- create log groups where delivery data will be stored by SNS
- subscribe these 2 log groups to the Lambda in charge of processing delivery receipts
- create SNS alarms in us-west-2 based on what we have in ca-central-1

I'll request spending limits in production and will add the production configuration after this PR has been merged and tested in staging

PR #128, opened 2021-01-06
fix: create SNS topics for alarms in us-west-2

AntoineAugusti:
Follow up of #126 

Turns out alarms require SNS topics to be in the same region, the [CI job failed](https://github.com/cds-snc/notification-terraform/runs/1658834964?check_suite_focus=true) when creating alarms.

> Creating metric alarm failed: ValidationError: Invalid region ca-central-1 specified. Only us-west-2 is supported.

This PR creates SNS topics for alarms in `us-west-2` and subscribes these topics to:
- Lambda functions for Slack webhooks
- Opsgenie in production

SNS topic subscriptions can be cross regions so for us this will be SNS topic used by alarms in `us-west-2` to Lambda functions (Slack integration) in `ca-central-1` + Opsgenie

> Amazon SNS supports the cross-region delivery of notifications to Amazon SQS queues and to AWS Lambda functions.
From https://docs.aws.amazon.com/sns/latest/dg/sns-cross-region-delivery.html

### DIFF
--- a/aws/common/sns.tf
+++ b/aws/common/sns.tf
@@ -19,8 +19,7 @@ resource "aws_sns_topic" "notification-canada-ca-alert-warning" {
 resource "aws_sns_topic" "notification-canada-ca-alert-warning-us-west-2" {
   provider = aws.us-west-2
 
-  name              = "alert-warning-us-west-2"
-  kms_master_key_id = aws_kms_key.notification-canada-ca.arn
+  name = "alert-warning-us-west-2"
 
   tags = {
     CostCenter = "notification-canada-ca-${var.env}"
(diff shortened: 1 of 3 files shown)


aws/common/sns.tf
```diff
@@ -16,6 +16,17 @@ resource "aws_sns_topic" "notification-canada-ca-alert-warning" {
   }
 }
 
+resource "aws_sns_topic" "notification-canada-ca-alert-warning-us-west-2" {
+  provider = aws.us-west-2
+
+  name              = "alert-warning-us-west-2"
+  kms_master_key_id = aws_kms_key.notification-canada-ca.arn
```
jimleroyer, line 23: I assume here we won't have a problem with KMS not being cross regions? 😓

jimleroyer (approved):
LGTM -- assuming that KMS will be cross-regional with these changes.

AntoineAugusti:
Thanks @jimleroyer!

Turns out KMS keys cannot be shared across regions.

> Q: What geographic region are my keys stored in?
> Keys generated by AWS KMS are only stored and used in the region in which they were created. They cannot be transferred to another region. For example; keys created in the EU-Central (Frankfurt) region are only stored and used within the EU-Central (Frankfurt) region.

Created a new one in `us-west-2` with reduced IAM permissions as it's only used by Cloudwatch alarms for now

PR #130, opened 2021-01-07
feat: forward cache-control header

AntoineAugusti:
I've set `Cache-Control: max-age=604800` (7 days) on files stored on the S3 static assets bucket (GoC logos, images on the homepage etc).

Forward this cache header at the Cloudfront level to tell clients to cache these images

Reference for syntax: https://stackoverflow.com/questions/63934583/what-is-the-syntax-for-headers-in-aws-cloudfront-distribution-forwarded-values

### DIFF
--- a/aws/cloudfront/cloudfront.tf
+++ b/aws/cloudfront/cloudfront.tf
@@ -25,6 +25,8 @@ resource "aws_cloudfront_distribution" "asset_bucket" {
     forwarded_values {
       query_string = false
 
+      headers = ["Cache-Control"]
+
       cookies {
         forward = "none"
       }


dorianjp:
@AntoineAugusti any concerns about asset caching for newly uploaded and then re-uploaded/changed assets such as custom headers?

AntoineAugusti:
@dorianjp Good question!

The header `Cache-Control: max-age=604800` needs to be applied manually (or we need to build something to do that). I've done it on the whole bucket in staging/prod now but ideally, we would need to do it sometimes for other assets.

The most important gain is for the GoC logos that are used often.

For reuploads, new images are assigned a temporary filename `temp-<hash>` so even if the cache header was applied/created automatically, this would not be an issue

AntoineAugusti:
Going to close this PR because I misunderstood `forwarded_values`. I thought it was Cloudfront forwarding an HTTP header to the client but this is the other way around.

This HTTP cache header is already forwarded to the end-user so we don't need to do any changes.

```
curl -I https://assets.notification.canada.ca/static/gov-canada-en.svg
HTTP/2 200 
content-type: image/svg+xml
content-length: 10048
date: Mon, 11 Jan 2021 15:22:06 GMT
last-modified: Fri, 08 Jan 2021 17:02:42 GMT
etag: "cc77f0718a88b1219088cd993501df54"
cache-control: max-age=604800
accept-ranges: bytes
server: AmazonS3
vary: Accept-Encoding
x-cache: Miss from cloudfront
via: 1.1 6bc96fcd042709dbc96cdeb6eb80307c.cloudfront.net (CloudFront)
x-amz-cf-pop: YUL62-C1
x-amz-cf-id: U190cbCrdvrKxckchAeIxkC0KL9Pr9OdIYk2qo8zdxx8PVpjgSZUTg==
```

PR #132, opened 2021-01-11
fix: specific source ARNs for Lambda permissions

AntoineAugusti:
Restrict invocation privileges for 2 Lambda functions to specific resources by specifying their ARNs.

From https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_permission, `source_arn`:
> When the principal is an AWS service, the ARN of the specific resource within that service to grant permission to. Without this, any resource from principal will be granted permission – even if that resource is from another account.

This PR keeps `principal` to AWS services (`(logs|sns).amazonaws.com` for us) but specifies ARNs for Cloudwatch log groups and SNS topics which are allowed to execute these Lambdas.

Remove unneeded `depends_on` statements which are now explicit because we cross-reference Lambda permissions/log groups or SNS topics so Terraform will be able to compute an execution plan.

### DIFF
--- a/aws/common/cloudwatch_log.tf
+++ b/aws/common/cloudwatch_log.tf
@@ -1,11 +1,6 @@
 resource "aws_cloudwatch_log_group" "sns_deliveries" {
   name = "sns/${var.region}/${var.account_id}/DirectPublishToPhoneNumber"
 
-  depends_on = [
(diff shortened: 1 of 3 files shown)


jimleroyer (approved):
LGTM

PR #133, opened 2021-01-11
feat: update period for SNS success rates' alarms

AntoineAugusti:
Previously these alarms were over a sliding window of 24 hours and we saw some issues over the weekend with low volumes.

Changing these alarms to:
- change the period from 24 hours to 12 hours
- trigger only if 2 consecutive periods of 12 hours are below the thresholds (a combination of `datapoints_to_alarm` and `evaluation_periods`)

This should be more robust to low volumes while avoiding a magic delay number or increasing the time it takes to detect a drop in success rate.

This changes 4 alarms: 1 warning and 1 critical alarm in `ca-central-1` (regular region) and the same thing in `us-west-2`(region used by the pool of dedicated long codes)


dorianjp (approved):
Good call on the re-sampling.

PR #134, opened 2021-01-11
fix: appropriate ARNs for log groups to Lambda

AntoineAugusti:
Follow up of #132.

According to https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_log_group for the `arn` output:

> The Amazon Resource Name (ARN) specifying the log group. Any `:*` suffix added by the API, denoting all CloudWatch Log Streams under the CloudWatch Log Group, is removed for greater compatibility with other AWS services that do not accept the suffix.

The suffix seems to be needed to allow the invocation of Lambda functions. This has the effect of allowing log messages to trigger to the Lambda and not the log group, which is indeed what we want.

This PR adds the required suffix to CloudWatch log groups ARNs.


jimleroyer (approved):
:shipit:

PR #147, opened 2021-01-15
feat: IAM config for continuous guardrail scanning

AntoineAugusti:
Create a new IAM role, user, credentials and policy required for continuous guardrail scanning.

https://github.com/cds-snc/continuous-guardrail-scanning


maxneuvians (approved):
Nice!

PR #148, opened 2021-01-18
feat: encrypt all S3 buckets

AntoineAugusti:
The continuous Guardrail scanning is not happy that we have [some S3 buckets without encryption enabled](https://github.com/cds-snc/continuous-guardrail-scanning/blob/a8c6a7051ef69f552db394ce3802b6f13c905fb3/239043911459/2021-01-16.txt#L263). These buckets hold images/logos so it was not _required_ to encrypt them but to follow the requirements and to please the CI scanning, I'm enabling encryption on these buckets.

To know more about the CI scanning #147


dorianjp (approved):
LGTM

PR #145, opened 2021-01-14
add new sns topic and slack integration

dorianjp:
https://trello.com/c/4e58AlOx/178-sns-warning-topic-should-only-be-used-by-alarms

* Added new topic "notification-canada-ca-alert-general"
* Re-assigned to new topic:
  * aws_health
  * rds
* Added new slack module "notify_slack_general"
  * bind to new general sns topic
  * new env vars for webhook and channel, copied values similar to existing slack integrations

PR #150, opened 2021-01-18
add new sns topic and slack integration pt2

### LINKED ISSUE #102: SNS warning topic should only be used by alarms
We have various [SNS topics defined](https://github.com/cds-snc/notification-terraform/blob/main/aws/common/sns.tf). Our main topics are used by alarms, to be notified when a warning or a critical event occurs.

These topics are used also by other components, which are not alarms. Example: the subscription to health events
https://github.com/cds-snc/notification-terraform/blob/bcbf0a715e4690e6da77ef1c9de4c47c728c462c/aws/common/cloudwatch_events.tf#L38

This has been set this way because the SNS topics are already [hooked up to Slack](https://github.com/cds-snc/notification-terraform/blob/main/aws/common/slack.tf) so it was convenient.

**Hooking up non-alarm events to SNS topics related to alarms is not a good practice and will mess up our statistics if we want to count the number of alarms down the road.**

Therefore, we should:
- review all things that use warning and critical SNS topics
- make sure that only alarms use these topics
- create new SNS topics for other events (no need to create one for each, maybe a generic one will do)
- make sure that these new topics are hooked up to Slack at least, consider sending these events to notifications-ops@cds-snc.ca as well

dorianjp:
closes #102 
Part 2 of #145 

Includes bindings for sns topic to inputs on terragrunt and slack module integration

-----
https://trello.com/c/4e58AlOx/178-sns-warning-topic-should-only-be-used-by-alarms

* Added new topic "notification-canada-ca-alert-general"
* Re-assigned to new topic:
  * aws_health
  * rds
* Added new slack module "notify_slack_general"
  * bind to new general sns topic
  * new env vars for webhook and channel, copied values similar to existing slack integrations

### DIFF
--- a/.github/workflows/merge_to_main_production.yml
+++ b/.github/workflows/merge_to_main_production.yml
@@ -27,10 +27,8 @@ env:
   TF_VAR_rds_cluster_password: ${{ secrets.PRODUCTION_RDS_CLUSTER_PASSWORD }}
   TF_VAR_cloudwatch_slack_webhook_warning_topic: ${{ secrets.PRODUCTION_CLOUDWATCH_SLACK_WEBHOOK }}
   TF_VAR_cloudwatch_slack_webhook_critical_topic: ${{ secrets.PRODUCTION_CLOUDWATCH_SLACK_WEBHOOK }}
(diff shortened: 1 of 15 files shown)


dorianjp:
[SNS warning Topic should only be used by Alarms](https://trello.com/c/4e58AlOx/178-sns-warning-topic-should-only-be-used-by-alarms)

PR #156, opened 2021-01-18
Enabled Athena with a dedicated S3 bucket

jimleroyer:
# Summary | Résumé

Let's have more tools to investigate issues around load balancers. AWS Athena can help us with that. [The trello card is here](https://trello.com/c/L775pbnO/243-enable-athena-engine-to-investigate-recent-load-balancers-issue).

We'll log into the AWS management console and see if AWS Athena is enabled and working with queries against load balancers.

This is a suggested checklist of questions reviewers might ask during their review | Voici une suggestion de liste de vérification comprenant des questions que les réviseurs pourraient poser pendant leur examen :

- [ ] Is it translated between both offical languages? | Est-ce dans les deux langues officielles?
- [ ] Does this cause automated test coverage to drop? | Est-ce que ça entraîne une baisse de la quantité de code couvert par les tests automatisés?
- [ ] Does this break existing functionality? | Est-ce que ça brise une fonctionnalité existante?
- [ ] Should this be split into smaller PRs to decrease change risk? | Est-ce que ça devrait être divisé en de plus petites demandes de tirage (« pull requests ») afin de réduire le risque lié aux modifications?
- [ ] Does this change the privacy policy? | Est-ce que ça entraîne une modification de la politique de confidentialité?
- [ ] Does this introduce any security concerns? | Est-ce que ça introduit des préoccupations liées à la sécurité?
- [ ] Does this significantly alter performance? | Est-ce que ça modifie de façon importante la performance?
- [ ] What is the risk level of using added dependencies? | Quel est le degré de risque d’utiliser des dépendances ajoutées?
- [ ] Should any documentation be updated as a result of this? (i.e. README setup, etc.) | Faudra-t-il mettre à jour la documentation à la suite de ce changement (fichier README, etc.)?

### DIFF
--- a/aws/common/athena.tf
+++ b/aws/common/athena.tf
@@ -2,8 +2,8 @@
 # Athena deployment configuration for the Notification application
 ###
 
-resource "aws_s3_bucket" "athena" {
-  bucket = "athena"
+resource "aws_s3_bucket" "athena_bucket" {
+  bucket = "notification-canada-ca-${var.env}-athena"
   acl    = "private"
   server_side_encryption_configuration {
     rule {
@@ -22,9 +22,48 @@ resource "aws_s3_bucket" "athena" {
     }
   }
 
+  logging {
+    target_bucket = aws_s3_bucket.athena_bucket_logs.bucket
+  }
+
   tags = {
     CostCenter = "notification-canada-ca-${var.env}"
   }
+}
+
+resource "aws_s3_bucket_public_access_block" "athena_bucket" {
+  bucket = aws_s3_bucket.athena_bucket.id
+
+  block_public_acls       = true
+  block_public_policy     = true
+  ignore_public_acls      = true
+  restrict_public_buckets = true
+}
+
+resource "aws_s3_bucket" "athena_bucket_logs" {
+  bucket = "notification-canada-ca-${var.env}-athena-logs"
+  acl    = "log-delivery-write"
+  server_side_encryption_configuration {
+    rule {
+      apply_server_side_encryption_by_default {
+        sse_algorithm = "AES256"
+      }
+    }
+  }
+
+  lifecycle_rule {
+    enabled = true
+
+    expiration {
+      days = 90
+    }
+  }
+
+  tags = {
+    CostCenter = "notification-canada-ca-${var.env}"
+  }
+
+  #tfsec:ignore:AWS002 - Ignore log of logs
 }
 
 resource "aws_athena_workgroup" "notification_athena_workgroup" {
@@ -35,7 +74,7 @@ resource "aws_athena_workgroup" "notification_athena_workgroup" {
     publish_cloudwatch_metrics_enabled = true
 
     result_configuration {
-      output_location = "s3://${aws_s3_bucket.athena.bucket}/output/"
+      output_location = "s3://${aws_s3_bucket.athena_bucket.bucket}/output/"
 
       encryption_configuration {
         encryption_option = "SSE_S3"
@@ -50,5 +89,5 @@ resource "aws_athena_workgroup" "notification_athena_workgroup" {
 
 resource "aws_athena_database" "notification_athena" {
   name   = "notification_athena"
-  bucket = aws_s3_bucket.athena.bucket
+  bucket = aws_s3_bucket.athena_bucket.bucket
 }
--- a/aws/common/athena.tf
+++ b/aws/common/athena.tf
@@ -9,6 +9,4 @@ resource "aws_athena_database" "notification_athena" {
   encryption_configuration {
     encryption_option = "SSE_S3"
   }
-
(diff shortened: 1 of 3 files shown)


jimleroyer:
[Enable Athena engine to investigate recent load balancers' issue](https://trello.com/c/L775pbnO/243-enable-athena-engine-to-investigate-recent-load-balancers-issue)

aws/common/athena.tf
```diff
@@ -0,0 +1,54 @@
...
+  name = "notification_athena_workgroup"
+
+  configuration {
+    enforce_workgroup_configuration    = true
+    publish_cloudwatch_metrics_enabled = true
+
+    result_configuration {
+      output_location = "s3://${aws_s3_bucket.athena.bucket}/output/"
+
+      encryption_configuration {
```
jimleroyer, line 40: I've put some encryption configuration and went with the easiest option to use:
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/athena_workgroup#encryption_configuration

Tell me if we want encryption and if that's the right option to use.

aws/common/athena.tf
```diff
@@ -0,0 +1,93 @@
...
+      days = 90
+    }
+  }
+
+  tags = {
+    CostCenter = "notification-canada-ca-${var.env}"
+  }
+
+  #tfsec:ignore:AWS002 - Ignore log of logs
+}
```
AntoineAugusti, line 67: Can you move S3 buckets to `s3.tf`?

aws/common/athena.tf
```diff
@@ -0,0 +1,93 @@
...
+      }
+    }
+  }
+
+  lifecycle_rule {
+    id      = "expire"
+    enabled = true
+
+    expiration {
+      days = 7
```
AntoineAugusti, line 21: 👍

aws/common/athena.tf
```diff
@@ -0,0 +1,93 @@
...
+        sse_algorithm = "AES256"
+      }
+    }
+  }
+
+  lifecycle_rule {
+    enabled = true
+
+    expiration {
+      days = 90
```
AntoineAugusti, line 58: 👍

aws/common/athena.tf
```diff
@@ -0,0 +1,93 @@
...
+      }
+    }
+  }
+
+  tags = {
+    CostCenter = "notification-canada-ca-${var.env}"
+  }
+}
+
+resource "aws_athena_database" "notification_athena" {
```
AntoineAugusti, line 90: Can you add encryption with `SSE_S3`?

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/athena_database#encryption_option

aws/common/athena.tf
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  tags = {
+    CostCenter = "notification-canada-ca-${var.env}"
+  }
+
+  #tfsec:ignore:AWS002 - Ignore log of logs
+}
+
+resource "aws_athena_workgroup" "notification_athena_workgroup" {
```
AntoineAugusti, line 69: We may be able to delete the workgroup. After [reading the documentation](https://docs.aws.amazon.com/athena/latest/ug/user-created-workgroups.html), we already have a primary workgroup we cannot delete and it seems like this is enough for us.

> By default, each account has a primary workgroup and the default permissions allow all authenticated users access to this workgroup. The primary workgroup cannot be deleted.

Can you take a look at the doc and confirm you understand the same thing?

AntoineAugusti (changes requested):

aws/common/athena.tf
jimleroyer: I will change to 7 to match with the previous one.

aws/common/athena.tf
jimleroyer: That makes sense.. It seems like workgroup is a feature added to have independent sandbox to manage queries, finance, quotas, permissions... we'd be good with the default in our case.

aws/common/athena.tf
```diff
@@ -0,0 +1,14 @@
...
+
+resource "aws_athena_database" "notification_athena" {
+  name   = "notification_athena"
+  bucket = aws_s3_bucket.athena_bucket.bucket
+
+  encryption_configuration {
+    encryption_option = "SSE_S3"
+  }
+
+  depends_on = [aws_s3_bucket.athena_bucket]
```
AntoineAugusti, line 13:
```suggestion
```
You should be able to leave this out as you're already referencing this resource in the bucket `field`

AntoineAugusti (approved):
Made a suggestion to delete an extra `depends_on`

LGTM, keep me posted if it works in staging

aws/common/athena.tf
jimleroyer: Ok got it

PR #163, opened 2021-01-22
feat: enable more log types on EKS

AntoineAugusti:
In order to understand better why some pods receive a `TERM` signal from Kubernetes while running, we need to have logs. This PR enables additional log types for our EKS cluster.

Relevant AWS documentation: https://docs.aws.amazon.com/eks/latest/userguide/control-plane-logs.html


dorianjp (approved):
log all the things

PR #164, opened 2021-01-22
fix: update EKS cluster log group name

AntoineAugusti:
Turns out that the Cloudwatch log group we have on Terraform is not the appropriate one. Therefore, we created an extra empty log group with a retention of 7 days.

This PR:
- fixes the EKS log group name
- bumps the retention from 7 days to 14 days to troubleshoot more easily

AWS documentation says
> Choose the cluster that you want to view logs for. The log group name format is `/aws/eks/<cluster-name>/cluster`.

https://docs.aws.amazon.com/eks/latest/userguide/control-plane-logs.html#viewing-control-plane-logs

## Screenshots from production

Note the empty log group with a retention of 7 days + the actual used one with 13 GB of logs

Screen Shot 2021-01-22 at 14 33 34
Screen Shot 2021-01-22 at 14 33 19

# Note
I think the `terragrunt apply` may fail because this log group already exists on AWS. If this is the case, I'll `terragrunt import` the resource to make sure that Terraform is aware of this existing resource and imports it in its state.

Did that in #157 already


aws/eks/variables.tf
```diff
@@ -53,3 +53,7 @@ variable "cloudfront_assets_arn" {
 variable "alb_log_bucket" {
   type = string
 }
+
+variable "eks_cluster_name" {
+  type = string
+}
```
AntoineAugusti, line 59: Had to introduce a variable because there is a dependency between the EKS cluster and the log group

env/staging/eks/terragrunt.hcl
```diff
@@ -67,6 +67,7 @@ inputs = {
   sns_alert_general_arn                  = dependency.common.outputs.sns_alert_general_arn
   alb_log_bucket                         = dependency.common.outputs.alb_log_bucket
   cloudfront_assets_arn                  = dependency.cloudfront.outputs.cloudfront_assets_arn
+  eks_cluster_name                       = "notification-canada-ca-staging-eks-cluster"
```
AntoineAugusti, line 70: Using `${var.env}` here does not work.

There is [a complicated solution](https://terragrunt.gruntwork.io/docs/features/locals/#including-globally-defined-locals) but I don't it's worth creating JSON/YAML env files and changing a lot of code to make it work.

dorianjp: is it ever worth it?

dorianjp (approved):
nice

AntoineAugusti:
Can confirm I had to import the resource first because it was externally managed before

Run locally
```
cd env/staging/eks
AWS_PROFILE=notify-staging terragrunt import aws_cloudwatch_log_group.notification-canada-ca-eks-cluster-logs "/aws/eks/notification-canada-ca-staging-eks-cluster/cluster"
```

And the CI job updated the retention as expected

PR #166, opened 2021-01-25
feat: update how AWS provider is declared

AntoineAugusti:
As part of [the Terraform upgrade](https://github.com/cds-snc/notification-terraform/pull/157#issuecomment-763047587), we need to change how providers are declared. This PR uses the new way, you can see warnings if you look at previous CI jobs about the old syntax.

Docs:
https://www.terraform.io/upgrade-guides/0-13.html#explicit-provider-source-locations
https://registry.terraform.io/providers/hashicorp/aws/latest/docs#example-usage

Example warning on https://github.com/cds-snc/notification-terraform/runs/1764494719?check_suite_focus=true
Screen Shot 2021-01-25 at 15 45 22

### DIFF
--- a/aws/common/athena.tf
+++ b/aws/common/athena.tf
@@ -1,5 +1,5 @@
 ###
-# Athena deployment configuration for the Notification application
+# Athena deployment configuration for Notify
 ###
 
 resource "aws_athena_database" "notification_athena" {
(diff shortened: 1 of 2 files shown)


aws/common/athena.tf
```diff
@@ -1,5 +1,5 @@
 ###
-# Athena deployment configuration for the Notification application
+# Athena deployment configuration for Notify
```
AntoineAugusti, line 2: @RenDavidDimen Did a fake commit to avoid being blocked as you saw in your recent README PR here

RenDavidDimen (approved):
Changes look good to me 👍

PR #167, opened 2021-01-29
feat: route delivery receipts to different queue

AntoineAugusti:
Route SES and SNS delivery receipts to a dedicated queue to not clog the Notify queue.

See the API PR for more details
https://github.com/cds-snc/notification-api/pull/1216

:warning: The API PR needs to be deployed before this one in staging and in production, otherwise no workers will pick up tasks from this queue.


AntoineAugusti:
Sorry for the commit/messages mess, `tfsec` acted weirdly for a few minutes and the Docker image was not pullable. Tried to fix that but in the end, I reverted to the version we use and now all is good.

Logged an issue to use the latest version thought #168

dorianjp (approved):
extra q

PR #169, opened 2021-01-29
feat: bump tfsec to 2.0.2 and fix violations

AntoineAugusti:
Fixes #168 

This PR updates [tfsec](https://github.com/tfsec/tfsec) to the latest version and addresses new security violations found:
- too old certificate on CloudFront
- not encrypting performance insights on RDS (requires a KMS key from `common`, added in #170)
- adding a web application firewall in front of assets served by CloudFront (does nothing at the moment but it'll be there and we could add rules)
- tfsec complained about [AWS051](https://tfsec.dev/docs/aws/AWS051/) (no encryption on RDS cluster) but it's safe to ignore as we're okay using AWS managed keys and our specific KMS key - checked with @maxneuvians

PR #170, opened 2021-01-29
feat: output the KMS's ARN

AntoineAugusti:
Output the KMS' ARN in order to use this value in another PR, #169.

Outputs should be available in the Terraform state beforehand, otherwise `terraform plan` will fail when adding a dependency between modules (the variable will not exist yet and Terraform will complain).


dorianjp (approved):
lgtm

PR #171, opened 2021-02-01
fix: update queue name for SES delivery receipts

AntoineAugusti:
Fixes #167 

Removed by mistake the `ca` part on the SES Lambda, I did not do the same mistake on the SNS Lambda.

https://github.com/cds-snc/notification-terraform/blob/fe02ba8d48a9ba6b52fb1cfc8fcc299adaa8f1a5/aws/common/lambdas/sns_to_sqs_sms_callbacks.py#L64

Discovered after performing tests in staging


dorianjp (approved):
we all fall down sometimes

PR #173, opened 2021-02-02
fix: don't use own KMS with RDS perf insights

AntoineAugusti:
Reworks #169

[CI failed](https://github.com/cds-snc/notification-terraform/actions/runs/530851929) when Terraform code was applied. We can't set a custom KMS key on Aurora instances if AWS keys were in use previously, at least through Terraform. Ignoring the error for now, it's fine to use the AWS managed encryption.


dorianjp (approved):
to move forward you must go back

PR #178, opened 2021-02-05
feat: create SES domain identity in us-east-1

AntoineAugusti:
Toward receiving emails with SES. This PR creates a domain identity in `us-east-1`, one of the 3 AWS regions where it's possible to receive emails.

After this has been merged, we'll need to verify the domain with DNS records. Notify DNS records [are managed elsewhere](https://github.com/cds-snc/dns/blob/master/terraform/notification.canada.ca-zone.tf) so I'll submit a PR there just after, to verify our domain and the next changes required.

Trello card: https://trello.com/c/PpAuHXEM/271-replies-to-notify-should-quickly-communicate-that-they-dont-work
AWS doc: https://docs.aws.amazon.com/ses/latest/DeveloperGuide/receiving-email-getting-started.html (this is step 2)

### DIFF
--- a/aws/dns/ses.tf
+++ b/aws/dns/ses.tf
@@ -32,6 +32,18 @@ resource "aws_ses_domain_mail_from" "notification-canada-ca" {
   mail_from_domain = "bounce.${aws_ses_domain_identity.notification-canada-ca.domain}"
 }
 
+###
+# Receiving emails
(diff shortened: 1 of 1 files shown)


jimleroyer (approved):
Domain looks good. :shipit:

PR #180, opened 2021-02-08
feat: add base Lambda for ses_receiving_emails

AntoineAugusti:
Follow up of https://github.com/cds-snc/notification-terraform/pull/178.

This PR is about creating a base Lambda, to handle SES incoming emails events after. We need to ship this first part now because we'll need to plug the Lambda ARN on SES, part of the `dns` module.

The next steps will be to:
- write a `ses_receipt_rule` to link SES to this Lambda https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ses_receipt_rule
- rework the Lambda function to post messages to a SQS queue that will trigger a Celery task on the API

### DIFF
--- a/aws/common/cloudwatch_log.tf
+++ b/aws/common/cloudwatch_log.tf
@@ -15,7 +15,7 @@ resource "aws_cloudwatch_log_group" "sns_deliveries_failures" {
 }
 
 resource "aws_cloudwatch_log_group" "ses_receiving_emails" {
-  name = "/aws/lambda/${aws_lambda_function.ses_receiving_emails.function_name}"
+  name = "/aws/lambda/${var.lambda_ses_receiving_emails_name}"
 
   retention_in_days = 90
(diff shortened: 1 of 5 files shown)


aws/common/cloudwatch_log.tf
```diff
@@ -14,6 +14,16 @@ resource "aws_cloudwatch_log_group" "sns_deliveries_failures" {
   }
 }
 
+resource "aws_cloudwatch_log_group" "ses_receiving_emails" {
```
AntoineAugusti, line 17: The Lambda can create its own log group but I'm creating it beforehand to be able to control the data retention (90 days) for logs.

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function#cloudwatch-logging-and-permissions

jimleroyer (approved):

PR #183, opened 2021-02-09
feat: SES receipt rule subscribe to Lambda

AntoineAugusti:
Follow up of https://github.com/cds-snc/notification-terraform/pull/180 and the fix in https://github.com/cds-snc/notification-terraform/pull/181

This PR is in charge of hooking SES to the previously created Lambda function, through a receipt rule.

Receipt rules have filters based on recipients (we handle all domains coming to our domain) and multiple actions. The only action we are interested in is handling the payload with a Lambda.

### DIFF
--- a/aws/dns/ses.tf
+++ b/aws/dns/ses.tf
@@ -45,6 +45,28 @@ resource "aws_ses_domain_identity" "notification-canada-ca-receiving" {
   domain = var.domain
 }
 
+resource "aws_ses_receipt_rule_set" "main" {
+  provider = aws.us-east-1
+
+  rule_set_name = "main"
+}
+
+resource "aws_ses_receipt_rule" "inbound-to-lambda" {
+  provider = aws.us-east-1
+
+  name          = "inbound-to-lambda"
+  rule_set_name = aws_ses_receipt_rule_set.main.rule_set_name
+  recipients    = [var.domain]
+  enabled       = true
+  scan_enabled  = true
+
+  lambda_action {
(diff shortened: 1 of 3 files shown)


aws/dns/ses.tf
```diff
@@ -45,6 +45,28 @@ resource "aws_ses_domain_identity" "notification-canada-ca-receiving" {
...
+}
+
+resource "aws_ses_receipt_rule" "inbound-to-lambda" {
+  provider = aws.us-east-1
+
+  name          = "inbound-to-lambda"
+  rule_set_name = aws_ses_receipt_rule_set.main.rule_set_name
+  recipients    = [var.domain]
+  enabled       = true
+  scan_enabled  = true
```
AntoineAugusti, line 61: > incoming emails will be scanned for spam and viruses

aws/dns/ses.tf
```diff
@@ -45,6 +45,28 @@ resource "aws_ses_domain_identity" "notification-canada-ca-receiving" {
...
+
+  lambda_action {
+    function_arn    = var.lambda_ses_receiving_emails_arn
+    invocation_type = "Event"
```
AntoineAugusti, line 65: We're not interested in a request/response, we're fine handling incoming emails asynchronously, after a few seconds.

https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html
> You can invoke a function synchronously (and wait for the response), or asynchronously. To invoke a function asynchronously, set InvocationType to Event.

jimleroyer (approved):
LGTM

PR #184, opened 2021-02-09
fix: apply Lambda permission for SES in us-east-1

AntoineAugusti:
You know what was missing from https://github.com/cds-snc/notification-terraform/pull/181?

Moving the `aws_lambda_permission` to `us-east-1` as well. This PR does just that.

Discovered when [production CI failed](https://github.com/cds-snc/notification-terraform/runs/1864632580?check_suite_focus=true), it was looking at the Lambda function in the wrong region.


jimleroyer:
I wish there was a real good IDEA dedicated to Terraform with plenty of documentation and static type checks.

jimleroyer (approved):
🤞

AntoineAugusti:
> I wish there was a real good IDEA dedicated to Terraform with plenty of documentation and static type checks.

Would love that. Surprises on apply are a nightmare.

PR #187, opened 2021-02-09
fix: appropriate Lambda permission for SES

AntoineAugusti:
SES was not allowed to invoke the Lambda function, this PR fixes it.

Turns out that `source_account` is a special parameter used by S3 and SES.
> This parameter is used for S3 and SES. The AWS account ID (without a hyphen) of the source owner

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_permission#source_account

The [SES documentation](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/receiving-email-permissions.html#receiving-email-permissions-lambda) confirms that the expected payload is

```json
{
  "Action": "lambda:InvokeFunction",
  "Principal": "ses.amazonaws.com",
  "SourceAccount": "111122223333",
  "StatementId": "GiveSESPermissionToInvokeFunction"
}
```

### DIFF
--- a/aws/common/lambda.tf
+++ b/aws/common/lambda.tf
@@ -144,5 +144,7 @@ resource "aws_lambda_permission" "ses_receiving_emails" {
   action         = "lambda:InvokeFunction"
   function_name  = aws_lambda_function.ses_receiving_emails.function_name
   principal      = "ses.amazonaws.com"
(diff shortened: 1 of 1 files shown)


jimleroyer (approved):
🤞

PR #188, opened 2021-02-10
feat: dispatch messages to SQS from SES inbound Lambda

AntoineAugusti:
Update the SES inbound Lambda to dispatch a message to SQS to trigger the Celery task introduced in https://github.com/cds-snc/notification-api/pull/1233

Still related to this [Trello card](https://trello.com/c/PpAuHXEM/271-replies-to-notify-should-quickly-communicate-that-they-dont-work) and PRs starting with #178

### DIFF
--- a/aws/common/lambda.tf
+++ b/aws/common/lambda.tf
@@ -63,7 +63,9 @@ resource "aws_lambda_function" "ses_receiving_emails" {
 
   environment {
     variables = {
+      CELERY_QUEUE_PREFIX   = var.celery_queue_prefix
       NOTIFY_SENDING_DOMAIN = var.domain
+      SQS_REGION            = var.region
     }
   }
 
@@ -141,9 +143,9 @@ resource "aws_lambda_permission" "sns_critical_us_west_2_to_slack_lambda" {
 resource "aws_lambda_permission" "ses_receiving_emails" {
   provider = aws.us-east-1
 
-  action         = "lambda:InvokeFunction"
-  function_name  = aws_lambda_function.ses_receiving_emails.function_name
-  principal      = "ses.amazonaws.com"
+  action        = "lambda:InvokeFunction"
+  function_name = aws_lambda_function.ses_receiving_emails.function_name
+  principal     = "ses.amazonaws.com"
   # tfsec:ignore:AWS058 Ensure that lambda function permission has a source arn specified
   # can ignore this because we specify `source_account` instead of `source_arn`
   source_account = var.account_id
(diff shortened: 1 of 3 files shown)


aws/common/lambdas/ses_receiving_emails.py
```diff
@@ -1,12 +1,62 @@
...
 from email.utils import parseaddr
 
+SENDING_DOMAIN = os.environ['NOTIFY_SENDING_DOMAIN']
+SQS_REGION = os.environ['SQS_REGION']
+CELERY_QUEUE_PREFIX = os.environ['SQS_REGION']
+CELERY_QUEUE = 'notify-internal-tasks'
+CELERY_TASK_NAME = 'send-notify-no-reply'
+
+
+def to_queue(queue, data):
```
AntoineAugusti, line 16: Logic is a standard copy-paste from other Lambas like here https://github.com/cds-snc/notification-terraform/blob/07be6d2ad5fed6081f32e5726c1fdca01ff04658/aws/common/lambdas/sns_to_sqs_sms_callbacks.py#L9

aws/common/lambda.tf
```diff
@@ -63,7 +63,9 @@ resource "aws_lambda_function" "ses_receiving_emails" {
 
   environment {
     variables = {
+      CELERY_QUEUE_PREFIX   = var.celery_queue_prefix
```
AntoineAugusti, line 66: A new addition, will reflect in other Lambdas as well to clean out Lambda code

aws/common/lambda.tf
```diff
@@ -63,7 +63,9 @@ resource "aws_lambda_function" "ses_receiving_emails" {
...
       NOTIFY_SENDING_DOMAIN = var.domain
+      SQS_REGION            = var.region
```
AntoineAugusti, line 68: Tried to use `AWS_REGION` but this is a reserved keyword. Anyway, it's important because the Lambda is in us-east-1 but the SQS queue in ca-central-1

aws/common/lambdas/ses_receiving_emails.py
```diff
@@ -1,12 +1,62 @@
...
+        "callbacks": None,
+        "errbacks": None,
+        "timelimit": [
+            None,
+            None
+        ],
+        "taskset": None,
+        "chord": None
+    }
+    print(f'Sending task {task}')
```
AntoineAugusti, line 35: print statements are logged to CloudWatch log, that's how we'll assess how much this feature is used

aws/common/lambdas/ses_receiving_emails.py
jimleroyer: Eventually, would it be easy to have this logic extracted into a library of its own so we can reuse?

jimleroyer (approved):
LGTM

aws/common/lambdas/ses_receiving_emails.py
AntoineAugusti: That's possible! We will need to look into including a shared Python file when building the ZIP archive required by Lambdas

PR #193, opened 2021-02-12
feat: prevent auto reply loops with inbound emails

AntoineAugusti:
Preventing "out of office/auto-reply loops" with the Notify auto-reply feature.

Some recipients reply to the Notify default auto-reply to the sender address and not the return path. Since the sender address is still a @notification.canada.ca, this email goes again through the same process. On and on. This is because we + the recipient implement poorly a defense mechanism to prevent this.

After looking at probematic payloads, we should be identify those in multiple ways:
1. the recipient is the GC Notify service `gc.notify.gc.notification@notification.canada.ca` in production
1. a `Precedence` header is set. This is not a real email standard but it seems used to identify automated responses/bulk emails
1. the subject contains specific words like "auto-reply"/"out of office" and the like

The PR relies on the `Precedence` header but we should look at other measures to prevent this in the future.

### DIFF
--- a/aws/common/lambdas/ses_receiving_emails.py
+++ b/aws/common/lambdas/ses_receiving_emails.py
@@ -9,6 +9,7 @@ from email.utils import parseaddr
 SENDING_DOMAIN = os.environ['NOTIFY_SENDING_DOMAIN']
 SQS_REGION = os.environ['SQS_REGION']
 CELERY_QUEUE_PREFIX = os.environ['CELERY_QUEUE_PREFIX']
+GC_NOTIFY_SERVICE_EMAIL = os.environ['GC_NOTIFY_SERVICE_EMAIL']
 CELERY_QUEUE = 'notify-internal-tasks'
 CELERY_TASK_NAME = 'send-notify-no-reply'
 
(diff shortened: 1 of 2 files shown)


AntoineAugusti:
Revised my initial PR to add another defense mechanism if the GC Notify is in the list of recipients. With this additional commit, we'll have measures 1 and 2 in place.

aws/common/lambdas/ses_receiving_emails.py
```diff
@@ -73,13 +78,22 @@ def lambda_handler(event, context):
     # https://docs.aws.amazon.com/ses/latest/DeveloperGuide/receiving-email-notifications-contents.html
     for record in event["Records"]:
         payload = record["ses"]
+        print(f"Full payload {payload}")
```
dorianjp, line 81: This is for debugging or another purpose?

dorianjp (approved):
Just the Q on the print statement.

aws/common/lambdas/ses_receiving_emails.py
AntoineAugusti: @dorianjp For debugging/logging purposes. I would not have been able to find a fix without the original payload before. Moved it to the top of the file because I want to see the payload even if the Lambda stops early because of a condition.

This log on the Lambda CloudWatch log group and it's deleted after 3 months. The payload does not contain the email body, just metadata.

jimleroyer (approved):
LGTM

PR #195, opened 2021-02-15
feat: set up default security group

AntoineAugusti:
Adresses https://github.com/cds-snc/site-reliability-engineering-internal/issues/104

> Ensure the default security group of every VPC restricts all traffic

Similar than https://github.com/cds-snc/covid-alert-server-staging-terraform/pull/136


maxneuvians (approved):
🚀

PR #201, opened 2021-02-24
Bump tfsec to 0.39.0

AntoineAugusti:
Upgrade tfsec and the underlying action. Tackled some new security checks and opened follow-up issues that will take more time or we need guidance from the SRE team.

#203 #202 

Did that now because #199 was blocked by tfsec which broke.

PR #204, opened 2021-02-24
fix: enable private access on EKS cluster

AntoineAugusti:
Follow up of #201, CI failed when applying

> Both private and public endpoint access cannot be false

Turned on private access, following the documentation https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_cluster#vpc_config

Was confused because I followed the "secure example" on [tfsec documentation](https://tfsec.dev/docs/aws/AWS069/) and this failed.


jimleroyer (approved):
LGTM

Issue #205, opened 2021-02-24
Tackle AWS069


AntoineAugusti:
See #201 #204

See documentation https://tfsec.dev/docs/aws/AWS069/

Merging this in staging removed our ability to use `kubectl`.

@maxneuvians, what would be an appropriate way to straighten up our k8s security will not be disturbing our operations?

PR #206, opened 2021-02-24
fix: revert EKS private access only

AntoineAugusti:
Reverts changes introduced in #201 and #204. Logged an issue to tackle this later #205


jimleroyer (approved):
Looks like a proper revert to me. 👍

PR #210, opened 2021-03-05
feat: create SES custom sending domains easily

AntoineAugusti:
**What / The story**

Notify offers the ability to send emails using the domain `notification.canada.ca` or a custom sending domain. Service owners can leverage a domain they own and use Notify to send emails.

At the moment we only have 1 service using this feature in production but more are coming - British Columbia wants to use a custom sending domain in the coming days. We need to streamline the process of letting them use custom sending domains and make it easier for the Notify team to do the required configuration changes.

**Why / The RoadMap link** BC gov raised this need in a Freshdesk ticket (attached)

**Done when / Output we want**

It's straightforward to add custom sending domains on our Terraform repository.

Trello card: https://trello.com/c/IU34UhxJ/359-create-custom-ses-sending-domains-easily

### DIFF
--- a/aws/dns/ses.tf
+++ b/aws/dns/ses.tf
@@ -74,7 +74,14 @@ resource "aws_ses_receipt_rule" "inbound-to-lambda" {
 }
 
 ###
-# Additional sending domains
+# Additional custom sending domains for emails
+# trvapply-vrtdemande.apps.cic.gc.ca is alone for historic reasons
+# and is not refactored to make sure the ressource is not destroyed/recreated.
+# Read the section "Refactoring Can Be Tricky"
+# https://blog.gruntwork.io/terraform-tips-tricks-loops-if-statements-and-gotchas-f739bbae55f9
+#
+# Afterwards there is a more automated way, using the list variable
+# `custom_sending_domains`.
 ###
 
 resource "aws_ses_domain_identity" "cic-trvapply-vrtdemande" {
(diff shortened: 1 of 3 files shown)


aws/dns/ses.tf
```diff
@@ -74,7 +74,14 @@ resource "aws_ses_receipt_rule" "inbound-to-lambda" {
 }
 
 ###
-# Additional sending domains
+# Additional custom sending domains for emails
+# trvapply-vrtdemande.apps.cic.gc.ca is alone for historic reasons
```
AntoineAugusti, line 78: Was done in https://github.com/cds-snc/notification-terraform/pull/89 in the past.

We don't want to clog this file forever, hence the addition of a set variable to manage custom sending domains from now on.

aws/dns/variables.tf
```diff
@@ -5,3 +5,7 @@ variable "notification_canada_ca_ses_callback_arn" {
 variable "lambda_ses_receiving_emails_arn" {
   type = string
 }
+
+variable "ses_custom_sending_domains" {
```
AntoineAugusti, line 9: This variable value will be different in staging/production. That's better than what was done in https://github.com/cds-snc/notification-terraform/pull/89, managing that with `count = var.env == "production" ? 1 : 0
`

aws/dns/ses.tf
```diff
@@ -111,3 +118,42 @@ resource "aws_ses_identity_notification_topic" "cic-trvapply-vrtdemande-complain
   include_original_headers = false
 }
 
+resource "aws_ses_domain_identity" "custom_sending_domains" {
+  for_each = var.ses_custom_sending_domains
```
AntoineAugusti, line 122: Using a set prevents from destroying/recreating resources. If resources are destroyed/created again, we risk disabling a domain but SES will regenerate DNS records to add - this could lead to the inability to send emails with custom domains, while clients do DNS changes.

Sets use the value as key+value and Terraform resources are indexed using that, making these resources resilient when changing the variable (adding/removing/changing domains).

Without a set, resources would be indexed using integers and resources could be destroyed/recreated
Screen Shot 2021-03-08 at 08 36 45

AntoineAugusti: @CalvinRodo used also this mechanism here! https://github.com/cds-snc/covid-alert-server-staging-terraform/blob/dcabfd71b8d0e086ba6284d4c294b6213b59e7c0/server/aws/ecr.tf#L1-L17

env/staging/dns/terragrunt.hcl
```diff
@@ -21,6 +21,7 @@ include {
 inputs = {
   notification_canada_ca_ses_callback_arn = dependency.common.outputs.notification_canada_ca_ses_callback_arn
   lambda_ses_receiving_emails_arn         = dependency.common.outputs.lambda_ses_receiving_emails_arn
+  ses_custom_sending_domains              = ["custom-sending-domain.staging.notification.cdssandbox.xyz"]
```
AntoineAugusti, line 24: We did not have a custom sending domain in staging, I'll use this PR as an opportunity to add one (I'll add the necessary DNS records after merging).

The end goal is to release this code to production and put `notification.gov.bc.ca` in the set.

Issue #162, opened 2021-01-22
Re create failure rate critical alarm for SMS in us-west-2 


AntoineAugusti:
This was temporarily deactivated in #160 and we need to bring it back after we've done the required work on the admin

Issue #106, opened 2020-12-10
Add alarms for throttled SMS queue


AntoineAugusti:
https://github.com/cds-snc/notification-api/pull/1185 introduced a new queue to handle throttled SMS. We already have alarms in place for the normal SMS queue.

Make sure that we have appropriate alarms for this new queue, at the very least something similar than
https://github.com/cds-snc/notification-terraform/blob/9be2da51eb926cfa9ddecc029415fb1ca07d4a4d/aws/common/cloudwatch_alarms.tf#L119-L133

AntoineAugusti:
Done in #213

PR #213, opened 2021-03-10
feat: rework SMS alarms

AntoineAugusti:
Our SMS alarms are lagging behind our SLOs and the fact that we now send SMS from multiple AWS regions using SNS.

Let's spend some time to:

- make sure that we have appropriate alarms in ca-central-1 and us-west-2
- reduce warning and critical thresholds for SMS
- add alarms on the throttled queue, used when sending SMS in us-west-2 https://github.com/cds-snc/notification-terraform/issues/106
- recreate a critical alarm we deleted in the past https://github.com/cds-snc/notification-terraform/issues/162

Trello card: https://trello.com/c/TdrBFem1/343-update-thresholds-for-sms-alarms

---

- reduced thresholds for SMS (checked CloudWatch data and we are far from these new thresholds but they are still way above our SLOs)
- refactored code
- add alarms on the throttled SMS queue https://github.com/cds-snc/notification-api/blob/4a0bc8e8847a47347d34fa70f0b6ee851387f8a2/app/notifications/process_notifications.py#L130-L133
- recreated critical alarm in `us-west-2` now that we send more SMS there

### DIFF
--- a/aws/common/cloudwatch_alarms.tf
+++ b/aws/common/cloudwatch_alarms.tf
@@ -119,6 +119,27 @@ resource "aws_cloudwatch_metric_alarm" "sns-sms-success-rate-canadian-numbers-cr
   }
 }
 
+resource "aws_cloudwatch_metric_alarm" "sns-sms-success-rate-canadian-numbers-us-west-2-critical" {
+  provider = aws.us-west-2
+
+  alarm_name          = "sns-sms-success-rate-canadian-numbers-us-west-2-critical"
+  alarm_description   = "SMS success rate to Canadian numbers is below 75% over 2 consecutive periods of 12 hours"
+  comparison_operator = "LessThanThreshold"
+  evaluation_periods  = "2"
+  datapoints_to_alarm = "2"
+  metric_name         = "SMSSuccessRate"
+  namespace           = "AWS/SNS"
+  period              = 60 * 60 * 12
+  statistic           = "Average"
+  threshold           = 75 / 100
+  alarm_actions       = [aws_sns_topic.notification-canada-ca-alert-critical-us-west-2.arn]
+  ok_actions          = [aws_sns_topic.notification-canada-ca-alert-critical-us-west-2.arn]
+  dimensions = {
+    SMSType = "Transactional"
+    Country = "CA"
+  }
+}
+
 resource "aws_cloudwatch_metric_alarm" "ses-bounce-rate-warning" {
   alarm_name          = "ses-bounce-rate-warning"
   alarm_description   = "Bounce rate >=5% over the last 12 hours"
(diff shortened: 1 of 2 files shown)


aws/common/cloudwatch_alarms.tf
```diff
@@ -175,34 +196,65 @@ resource "aws_cloudwatch_metric_alarm" "ses-complaint-rate-critical" {
...
+  alarm_description   = "ApproximateAgeOfOldestMessage in SMS queue is older than 1 minute in a 5-minute period"
   comparison_operator = "GreaterThanOrEqualToThreshold"
-  evaluation_periods  = "3"
   metric_name         = "ApproximateAgeOfOldestMessage"
   namespace           = "AWS/SQS"
-  period              = "300"
-  extended_statistic  = "p90"
-  threshold           = 60 * 3
+  period              = 60 * 5
+  statistic           = "Average"
```
AntoineAugusti, line 205: p90 is not a valid statictic for this metric https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-available-cloudwatch-metrics.html#sqs-metrics

aws/common/cloudwatch_alarms.tf
```diff
@@ -119,6 +119,27 @@ resource "aws_cloudwatch_metric_alarm" "sns-sms-success-rate-canadian-numbers-cr
...
+  period              = 60 * 60 * 12
+  statistic           = "Average"
+  threshold           = 75 / 100
+  alarm_actions       = [aws_sns_topic.notification-canada-ca-alert-critical-us-west-2.arn]
+  ok_actions          = [aws_sns_topic.notification-canada-ca-alert-critical-us-west-2.arn]
+  dimensions = {
+    SMSType = "Transactional"
+    Country = "CA"
+  }
+}
```
AntoineAugusti, line 141: A strict revert of https://github.com/cds-snc/notification-terraform/issues/162

AntoineAugusti:
I will update the spreadsheet after a review

aws/common/cloudwatch_alarms.tf
```diff
@@ -175,34 +196,67 @@ resource "aws_cloudwatch_metric_alarm" "ses-complaint-rate-critical" {
...
+  alarm_description   = "ApproximateAgeOfOldestMessage in SMS queue is older than 1 minute for 10 minutes"
   comparison_operator = "GreaterThanOrEqualToThreshold"
-  evaluation_periods  = "3"
+  evaluation_periods  = "2"
   metric_name         = "ApproximateAgeOfOldestMessage"
   namespace           = "AWS/SQS"
-  period              = "300"
-  extended_statistic  = "p90"
-  threshold           = 60 * 5
+  period              = 60 * 5
```
dorianjp, line 220: With the averaging statistic on, we are taking the 2 periods together to average above the threshold collectively, correct? the averaging is based on hits above the 60 second age mark?

dorianjp (approved):
Just some clarifying questions, but it looks fine.

aws/common/cloudwatch_alarms.tf
AntoineAugusti: We can't use "Percentile-Based CloudWatch Alarms and Low Data Samples" unfortunately because the `SMSSuccessRate` [valid statistics](https://docs.aws.amazon.com/sns/latest/dg/sns-monitoring-using-cloudwatch.html) are only `Sum, Average, Data Samples`. Because percentiles are out, this solution is out.

We could look at [Composite alarms](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Create_Composite_Alarm.html). We would create another alarm, counting the number of notifications in a period and disable this success rate if the number is too low for us.

Do you want to go this route now or a bit later?

aws/common/cloudwatch_alarms.tf
AntoineAugusti: This alarm computes the average of `ApproximateAgeOfOldestMessage` over 2 periods of 5 minutes. To be above 60s in a 5-minute period, it means that a message stayed more than 60s in the queue without being picked up, on average, during 2 consecutive periods of 5 minutes. This may not be the same message and a single message breaching this is enough for the alarm to trigger.

You'll see that the situation would be way worse than our [SLO for SMS](https://docs.google.com/spreadsheets/d/1fU-FJ7THfWEqNhbpQ4r22MQipFwC7SP851ZQnOf68cM/edit#gid=0), where we aim to send SMS:
- 90% < 10s
- 99% < 60s

Screen Shot 2021-03-16 at 13 56 27

AntoineAugusti:
Thanks @dorianjp for your careful review. After chatting on Slack, we made our due diligence looking at CloudWatch metrics data. Added `treat_missing_data = "notBreaching"` in the latest commit to be extra cautious. We're fine not alerting if we have no SMS traffic over the night/weekend.

PR #214, opened 2021-03-11
feat: remove encryption for SNS topic used by RDS events

AntoineAugusti:
RDS event notifications should be delivered to [**unencrypted** SNS topics](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_Events.html#USER_Events.Subscribing). Our current SNS topic is encrypted so we missed a few database events.

Fixing that to make sure we receive database notifications.

See AWS case support in staging

> [Case 8080942541] Missing notification for cluster event subscription

Related to https://github.com/cds-snc/notification-terraform/pull/145

### DIFF
--- a/aws/common/sns.tf
+++ b/aws/common/sns.tf
@@ -52,7 +52,7 @@ resource "aws_sns_topic" "notification-canada-ca-alert-critical-us-west-2" {
 resource "aws_sns_topic" "notification-canada-ca-alert-general" {
   name = "alert-general"
 
-  # tfsec:ignore:AWS016
+  # tfsec:ignore:AWS016 Unencrypted SNS topic
   # > Amazon RDS event notification is only available for unencrypted SNS topics.
   # > If you specify an encrypted SNS topic, event notifications aren't sent for the topic.
   #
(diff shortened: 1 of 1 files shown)


aws/common/sns.tf
```diff
@@ -50,9 +50,13 @@ resource "aws_sns_topic" "notification-canada-ca-alert-critical-us-west-2" {
 }
 
 resource "aws_sns_topic" "notification-canada-ca-alert-general" {
-  name              = "alert-general"
-  kms_master_key_id = aws_kms_key.notification-canada-ca.arn
+  name = "alert-general"
```
AntoineAugusti, line 53: Checked that this SNS topic is not used by other things. Messages that will go through will not contain PII, only information like "a database upgrade is available"/"this database was restarted" etc

dorianjp: So all we need to do is remove the key id?
I took a peek at #22 and see these have always been setup w/encryption.

dorianjp (approved):
Some clarification, otherwise good.

AntoineAugusti:
@dorianjp AWS support wasn't sure if we needed to delete/add again the SNS subscription in RDV events on top of removing the encryption on the SNS channel.

I don't see why this would be needed: the subscription is marked as active in the AWS console. My feeling is that an external system cannot add things to this topic because it doesn't have the appropriate permissions to encrypt payloads.

I'll merge this PR, reboot a database in staging and see if we've got a notification.

PR #224, opened 2021-03-23
feat: add lifecycle rule for attachments

AntoineAugusti:
Expire files stored in the document download bucket after 3 days if they are stored in the `tmp/` folder.

Trello: https://trello.com/c/wtkWVPcA/369-let-clients-attach-files-directly-to-emails-with-the-api


dorianjp (approved):
looks like the right bucket

AntoineAugusti:
✅ 
Screen Shot 2021-03-24 at 07 54 15

PR #227, opened 2021-03-29
Add DescribeDBClusters permission for inspec

AntoineAugusti:
Add the `rds:DescribeDBClusters` permission to the `inspec` user, running [continuous guardrails scanning](https://github.com/cds-snc/continuous-guardrail-scanning).

Will send a PR soon to the other repository, listing database clusters and making sure that deletion protection is always enabled. Submitting this PR in advance to make sure we have the required permissions when the CI will be updated.

This is an incident action item.

PR #228, opened 2021-03-30
Bump production to 0.85.0

AntoineAugusti:
Releases https://github.com/cds-snc/notification-terraform/pull/227 to production.

Confirmed in staging that the new permission has been added


QuanNguyen-CDS (approved):
Good: checked on staging and the permissions for groups > inspec are good!

PR #233, opened 2021-03-30
Bump production to 0.86.1

AntoineAugusti:
Release #229 #232 to production, set the threshold for production.


QuanNguyen-CDS (approved):
yeah looks good on staging

PR #234, opened 2021-03-31
prevent destroying RDS cluster instance

QuanNguyen-CDS:
# Summary | Résumé
[Trello](https://trello.com/c/bRRjSwpl)
feat: added prevent_destroy=true within the RDS cluster lifecycle define
fix: added prevent_destroy=true within the rds cluster instance lifecycle define

Validate the RDS plan does not adversely change the configuration


AntoineAugusti (changes requested):
My bad, the Trello card was poorly written. This change is fine and it's useful but it's important to put it on [RDS cluster instances](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/rds_cluster_instance). Those resources were destroyed last time, not the cluster.

AntoineAugusti (approved):

PR #236, opened 2021-04-06
feat: add CloudWatch queries

AntoineAugusti:
`aws_cloudwatch_query_definition` is a new resource available on the AWS provider https://github.com/hashicorp/terraform-provider-aws/issues/13223

We had a bunch of queries saved in production, those have been created using the AWS console. Recreate our queries in Terraform to make sure that we have the same queries between staging and production.

Trello: https://trello.com/c/MMknwMKW/420-add-cloudwatch-queries-on-terraform

### DIFF
--- a/aws/eks/cloudwatch_queries.tf
+++ b/aws/eks/cloudwatch_queries.tf
@@ -38,7 +38,7 @@ resource "aws_cloudwatch_query_definition" "services-over-daily-rate-limit" {
   ]
 
   query_string = <<QUERY
-fields @timestamp, log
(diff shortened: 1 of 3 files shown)


aws/eks/cloudwatch_queries.tf
```diff
@@ -0,0 +1,46 @@
+resource "aws_cloudwatch_query_definition" "admin-api-50X-errors" {
```
AntoineAugusti, line 1: Those have been copy/pasted from the AWS console in production

PR #250, opened 2021-04-21
feat: update Celery errors query to catch various workers

AntoineAugusti:
Reflect changes introduced in https://github.com/cds-snc/notification-terraform/pull/247 in the CloudWatch query


QuanNguyen-CDS (approved):
Good

PR #251, opened 2021-04-21
feat: handle no recipients for SES inbound

AntoineAugusti:
We've seen payloads in production where To + Cc + Bcc fields are not present in `commonHeaders` but `recipients` is filled from the receipt.

`recipients` is defined as:

> A list of recipients (specifically, the envelope RCPT TO addresses) that were matched by the active receipt rule. The addresses listed here may differ from those listed by the destination field in the mail object.

Use this new field to a build a complete list of recipients

AWS doc: https://docs.aws.amazon.com/ses/latest/DeveloperGuide/receiving-email-notifications-contents.html
Difference between RCPT TO and TO: https://stackoverflow.com/questions/10822190/in-smtp-must-the-rcpt-to-and-to-match

PS: for Notify developers, I can share JSON payload on request.

### DIFF
--- a/aws/common/lambdas/ses_receiving_emails.py
+++ b/aws/common/lambdas/ses_receiving_emails.py
@@ -59,7 +59,7 @@ def to_queue(queue, data):
 
 def parse_recipients(headers, receipt):
     # Gather recipients from our own domain only
-    recipients = headers.get("to", []) + headers.get("cc", []) + headers.get("bcc", []) + receipt.get('recipients', [])
+    recipients = headers.get("to", []) + headers.get("cc", []) + headers.get("bcc", []) + receipt.get("recipients", [])
     recipients = [parseaddr(r)[1] for r in set(recipients)]
 
     return [r for r in recipients if r.endswith(f"@{SENDING_DOMAIN}")]
--- a/aws/common/lambdas/ses_receiving_emails.py
+++ b/aws/common/lambdas/ses_receiving_emails.py
@@ -60,7 +60,7 @@ def to_queue(queue, data):
 def parse_recipients(headers, receipt):
     # Gather recipients from our own domain only
     recipients = headers.get("to", []) + headers.get("cc", []) + headers.get("bcc", []) + receipt.get("recipients", [])
-    recipients = [parseaddr(r)[1] for r in set(recipients)]
+    recipients = [parseaddr(r)[1] for r in set(r.lower() for r in recipients)]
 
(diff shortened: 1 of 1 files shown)


aws/common/lambdas/ses_receiving_emails.py
```diff
@@ -57,10 +57,10 @@ def to_queue(queue, data):
     queue.send_message(MessageBody=msg)
 
 
-def parse_recipients(headers):
+def parse_recipients(headers, receipt):
     # Gather recipients from our own domain only
-    recipients = headers.get("to", []) + headers.get("cc", []) + headers.get("bcc", [])
-    recipients = [parseaddr(r)[1] for r in recipients]
+    recipients = headers.get("to", []) + headers.get("cc", []) + headers.get("bcc", []) + receipt.get('recipients', [])
+    recipients = [parseaddr(r)[1] for r in set(recipients)]
```
AntoineAugusti, line 63: The `set` should have been here already

QuanNguyen-CDS: i would suggest that we str.lower() everything before we do the set. Emails are not case sensitive.
I'm seeing with the [RFC 5321](https://tools.ietf.org/html/rfc5321#section-4.5.3.2.3) cases where the RCPT TO: envelope is different than the To: address however ultimately have the same email address

```
      S: 220 foo.com Simple Mail Transfer Service Ready
      C: EHLO bar.com
      S: 250-foo.com greets bar.com
      S: 250-8BITMIME
      S: 250-SIZE
      S: 250-DSN
      S: 250 HELP
      C: MAIL FROM:<JQP@bar.com>
      S: 250 OK
      C: RCPT TO:<Jones@XYZ.COM>
      S: 250 OK
      C: DATA
      S: 354 Start mail input; end with <CRLF>.<CRLF>
      C: Date: Thu, 21 May 1998 05:33:29 -0700
      C: From: John Q. Public <JQP@bar.com>
      C: Subject: The Next Meeting of the Board
      C: To: Jones@xyz.com
      C:
      C: Bill:
      C: The next meeting of the board of directors will be
      C: on Tuesday.
      C: John.
      C: .
      S: 250 OK
      C: QUIT
      S: 221 foo.com Service closing transmission channel
```

AntoineAugusti: Sounds good

QuanNguyen-CDS (approved):
Nice, thank you

PR #246, opened 2021-04-19
feat: specify SSL policy for ALB

AntoineAugusti:
Pick an appropriate SSL policy for the load balancer, according to:

- recommendations: https://canada-ca.github.io/cloud-guardrails/EN/07_Protect-Data-in-Transit.html
- what other teams do: https://github.com/cds-snc/covid-alert-server-staging-terraform/blob/6232ad547cba9d4dd57300ea4377b2e3821442e7/server/aws/lb.tf#L82

See https://aws.amazon.com/about-aws/whats-new/2019/10/application-load-balancer-and-network-load-balancer-add-new-security-policies-for-forward-secrecy-with-more-strigent-protocols-and-ciphers/

PR #262, opened 2021-05-10
feat: add listener for legacy TLS

AntoineAugusti:
At the end of June 2021, Notify will only support TLS 1.2. Some clients cannot meet the deadline and their API integration would be affected.

This PR adds a listener to our load balancer that will keep using the legacy TLS setting. This endpoint will be used by clients that cannot upgrade for now. This PR also adds a new firewall rule to allow traffic on this new port.

See #246 #253


aws/eks/securitygroups.tf
```diff
@@ -21,6 +21,13 @@ resource "aws_security_group" "notification-canada-ca-alb" {
     cidr_blocks = ["0.0.0.0/0"] #tfsec:ignore:AWS008
   }
 
+  ingress {
+    protocol    = "tcp"
+    from_port   = 4444
+    to_port     = 4444
+    cidr_blocks = ["0.0.0.0/0"] #tfsec:ignore:AWS008
+  }
```
jimleroyer, line 29: Which 4444 port does this redirect to? `4444` to the app? (I assume not) 🤔

jimleroyer (approved):

aws/eks/securitygroups.tf
AntoineAugusti: Clarified on Slack how load balancer/target groups and app work!

PR #263, opened 2021-05-10
feat: update healthcheck alarms' unit

AntoineAugusti:
The value was not reported in milliseconds on StatsD in the past, this PR fixed the bug https://github.com/cds-snc/notification-utils/pull/82#discussion_r625875476 and now this alarm needs to be updated.

### DIFF
--- a/aws/common/cloudwatch_alarms.tf
+++ b/aws/common/cloudwatch_alarms.tf
@@ -401,7 +401,7 @@ resource "aws_cloudwatch_metric_alarm" "healtheck-page-slow-response-warning" {
   namespace           = "NotificationCanadaCa"
   period              = "300"
   statistic           = "Average"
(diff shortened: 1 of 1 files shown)


aws/common/cloudwatch_alarms.tf
```diff
@@ -401,7 +401,7 @@ resource "aws_cloudwatch_metric_alarm" "healtheck-page-slow-response-warning" {
   namespace           = "NotificationCanadaCa"
   period              = "300"
   statistic           = "Average"
-  threshold           = 1
+  threshold           = 100
```
AntoineAugusti, line 404: Matches the description "Healthcheck page response time is above 100ms for 10 minutes"

jimleroyer (approved):
🚢  it

PR #264, opened 2021-05-10
Bump production to 0.97.0

AntoineAugusti:
Release https://github.com/cds-snc/notification-terraform/pull/262 (may need more work but doesn't break existing features) and #263


AntoineAugusti:
Confirmed #262 but some ISP may block this port, let's see if our client has troubles with this port and if yes we will change that

PR #266, opened 2021-05-18
Log DNS queries made in our VPC

jimleroyer:
# Summary | Résumé

* Created a new log group to store the DNS requests made within our VPC.
* Created a Resolver query logging within Route 53 to log DNS requests.

I tried to tests first with CLI but I didn't have access to latest documented commands despite having latest available (at least from Brew), such as `create-resolver-query-log-config` and `associate-resolver-query-log-config`.

For now, we'll have to test by applying Terraform changes when merging into the `main` branch.

      langues officielles?
      fonctionnalité existante?

### DIFF
--- a/aws/common/cloudwatch_log.tf
+++ b/aws/common/cloudwatch_log.tf
@@ -50,8 +50,6 @@ resource "aws_cloudwatch_log_group" "sns_deliveries_failures_us_west_2" {
 }
 
 resource "aws_cloudwatch_log_group" "route53_resolver_query_log" {
-  provider = aws.us-east-1
-
   name = "route53/${var.region}/${var.account_id}/DNS/logs"
 
   retention_in_days = 14
--- a/aws/common/cloudwatch_log.tf
+++ b/aws/common/cloudwatch_log.tf
@@ -49,6 +49,18 @@ resource "aws_cloudwatch_log_group" "sns_deliveries_failures_us_west_2" {
   }
 }
 
+resource "aws_cloudwatch_log_group" "route53_resolver_query_log" {
+  provider = aws.us-east-1
+
+  name = "route53/${var.region}/${var.account_id}/DNS/logs"
+
+  retention_in_days = 14
+
+  tags = {
+    CostCenter = "notification-canada-ca-${var.env}"
+  }
+}
+
 ###
 # AWS CloudWatch Logs Metrics
 ###
--- a/aws/common/route53.tf
+++ b/aws/common/route53.tf
@@ -1,0 +1,12 @@
+###
+# AWS Route 53 for Notification application
+###
+
+resource "aws_route53_resolver_query_log_config" "dns_query_log_config" {
+  name            = "us-west-2/${var.account_id}/dns_query_log_config"
+  destination_arn = [aws_cloudwatch_log_group.route53_resolver_query_log.arn]
+
+  tags = {
+    CostCenter = "notification-canada-ca-${var.env}"
+  }
+}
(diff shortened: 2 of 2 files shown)


aws/common/route53.tf
```diff
@@ -0,0 +1,12 @@
+###
+# AWS Route 53 for Notification application
+###
+
+resource "aws_route53_resolver_query_log_config" "dns_query_log_config" {
+  name            = "us-east-1/${var.account_id}/dns_query_log_config"
+  destination_arn = [aws_cloudwatch_log_group.route53_resolver_query_log.arn]
```
AntoineAugusti, line 7: This should be a string

aws/common/cloudwatch_log.tf
```diff
@@ -49,6 +49,18 @@ resource "aws_cloudwatch_log_group" "sns_deliveries_failures_us_west_2" {
   }
 }
 
+resource "aws_cloudwatch_log_group" "route53_resolver_query_log" {
+  provider = aws.us-east-1
```
AntoineAugusti, line 53: I think that you read the wrong documentation page. [This page](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/query-logs.html) is about query logging when you use Route53 as a DNS resolver for your own domain that you're managing on AWS.

This is not what we're using, we're using their resolver for everything in our VPC.

You can visit https://ca-central-1.console.aws.amazon.com/route53resolver/home?region=ca-central-1#/query-logging in our staging account to see that we could be able to use a log group in ca-central-1

aws/common/route53.tf
```diff
@@ -0,0 +1,12 @@
+###
+# AWS Route 53 for Notification application
+###
+
+resource "aws_route53_resolver_query_log_config" "dns_query_log_config" {
```
AntoineAugusti, line 5: Needs also a https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route53_resolver_query_log_config_association to associate this with our VPC!

aws/common/route53.tf
```diff
@@ -0,0 +1,12 @@
...
+  name            = "us-east-1/${var.account_id}/dns_query_log_config"
```
AntoineAugusti, line 6: Change the name as this will not be in this region. You can use `var.region`

AntoineAugusti (changes requested):

aws/common/route53.tf
jimleroyer: I did notice from the same message given by the build. Other examples in the code does not enclose in strings though for similarly ARN references. Should we simply enclose it in double quotes and that will be ok?

AntoineAugusti: I think you don't need to enclose it in double quotes! `make fmt` would convert that as needed

aws/common/cloudwatch_log.tf
jimleroyer: Sorry, I don't understand your comment. A log group is defined here to be used later by the aws route 53 resolver query log. The link you referred to is to create a Route 53 resolver query log too. 

Let's speak on Slack about this.

jimleroyer: Spoke with Antoine offline. He meant to say that the provider's region is incorrect.

aws/common/route53.tf
jimleroyer: I was 😫 🍸

AntoineAugusti (approved):
:shipit: 

just after you fix the name of the resolver query log config

jimleroyer:
> :shipit:
> 
> just after you fix the name of the resolver query log config

The name was changed to `name            = "${var.region}/${var.account_id}/dns_query_log_config"` in my last commit. Do you suggest something else?

PR #241, opened 2021-04-12
feat: detect when SMS are marked as spam or carrier is unavailable

AntoineAugusti:
Add log metric filters to detect when:
- SMS are marked as spam by a phone carrier
- when a phone carrier is unavailable

from SNS SMS log groups, in 2 AWS regions where we send SMS.

Add **warning alarms** using these new metrics:
- More than 10 SMS have been blocked as spam over 12 hours
- More than 10 SMS failed because a phone carrier is unavailable over 6 hours

Trello: https://trello.com/c/wzAeiXeO/365-be-alerted-when-phone-carriers-are-having-issues

PR #274, opened 2021-06-07
feat: add alarms for throttled SMS

AntoineAugusti:
Add alarms when:
- we get a `Rate exceeded.` error from SNS SMS in us-west-2 or ca-central-1. This happens when we send more than 1 SMS per second per long code phone number
- we consume too many message per minute from the `send-throttled-sms-tasks` SQS queue, where we are not supposed to consume more than 1 message per second

These alarms are warning alarms only, but they will be useful to ensure that our application is running fine and may detect bugs.

### DIFF
--- a/aws/common/cloudwatch_alarms.tf
+++ b/aws/common/cloudwatch_alarms.tf
@@ -204,6 +204,36 @@ resource "aws_cloudwatch_metric_alarm" "sns-sms-phone-carrier-unavailable-us-wes
   treat_missing_data  = "notBreaching"
 }
 
+resource "aws_cloudwatch_metric_alarm" "sns-sms-rate-exceeded-warning" {
+  alarm_name          = "sns-sms-rate-exceeded-warning"
+  alarm_description   = "At least 1 SNS SMS rate exceeded error in 5 minutes"
+  comparison_operator = "GreaterThanOrEqualToThreshold"
+  evaluation_periods  = "1"
+  metric_name         = aws_cloudwatch_log_metric_filter.sns-sms-rate-exceeded.metric_transformation[0].name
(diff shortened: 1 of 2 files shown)


aws/common/cloudwatch_log.tf
```diff
@@ -122,6 +122,40 @@ resource "aws_cloudwatch_log_metric_filter" "sns-sms-phone-carrier-unavailable-u
   }
 }
 
+resource "aws_cloudwatch_log_metric_filter" "sns-sms-rate-exceeded" {
```
AntoineAugusti, line 125: If you look above, you will see very similar code for spam/carrier unavailable https://github.com/cds-snc/notification-terraform/pull/241

aws/common/cloudwatch_alarms.tf
```diff
@@ -392,6 +422,25 @@ resource "aws_cloudwatch_metric_alarm" "sqs-bulk-queue-delay-critical" {
...
+  alarm_description   = "NumberOfMessagesReceived is more than the expected maximum rate for send-throttled-sms-tasks SQS queue"
+  comparison_operator = "GreaterThanOrEqualToThreshold"
+  evaluation_periods  = "1"
+  metric_name         = "NumberOfMessagesReceived"
+  namespace           = "AWS/SQS"
+  period              = 60
+  statistic           = "Sum"
+  # Maximum rate is supposed to be 1 SMS per second, so 60 messages
+  # per minute, but giving it a 10% room.
+  threshold     = 60 * 1.1
```
AntoineAugusti, line 436: Saw that in production we sometimes have 61 or 62 messages per 60s over a period of 4 weeks.

jimleroyer (approved):

PR #289, opened 2021-08-18
feat: add terraform for aws lambda deployment of api

smcmurtry:
# Summary | Résumé

So far I have borrowed some code that looked relevant from here: https://github.com/cds-snc/scan-websites/tree/main/terragrunt/aws/api

I'm not sure what I'm doing so I am putting this up as a draft for now.


aws/common/lambda.tf
```diff
@@ -77,6 +77,65 @@ resource "aws_lambda_function" "ses_receiving_emails" {
...
+  image_uri    = "${aws_ecr_repository.api.repository_url}:latest"
+
+  role    = aws_iam_role.api.arn
+  timeout = 60
+
+  memory_size = 1024
+
+  environment {
+    variables = {
+      ADMIN_CLIENT_SECRET =
```
smcmurtry, line 93: Do I need to create an inputs.tf file (like [this](https://github.com/cds-snc/scan-websites/blob/main/terragrunt/aws/api/inputs.tf)) for all of these environment variables?

smcmurtry: Okay I think I should just add entries to aws/common/variables.tf (or perhaps move things to a new folder aws/lambda/?) like this:
```
variable "admin_client_secret" {
  type = string
  sensitive   = true
}
```

patheard: Yup, that's correct - Terraform will look in a few places to try and [assign variable values](https://www.terraform.io/docs/language/values/variables.html#assigning-values-to-root-module-variables).  In the case of your pipeline, it's using the `TF_VAR_` environment var method.

So in your case, you'd just need to make sure there was a `TF_VAR_admin_client_secret` in scope and the value would get set during the `terraform plan/apply` commands.

One other note about needing an `inputs.tf` file.  That's just convention around the name.  You could add the `variable` definition block to any `.tf` file that's in the same Terraform module and it would get correctly recognized as a variable.

aws/common/ecr.tf
```diff
@@ -0,0 +1,10 @@
+resource "aws_ecr_repository" "api" {
+  # checkov:skip=CKV_AWS_51:The :latest tag is used in Staging
```
patheard, line 2: This will need to change to a tfsec style ignore comment:
```
# tfsec:ignore:AWS078
```
We're using [checkov](https://github.com/bridgecrewio/checkov) on the scan-websites project which is why we've got a different ignore comment style.

aws/common/api_gateway.tf
```diff
@@ -0,0 +1,74 @@
...
+  lifecycle {
+    create_before_destroy = true
+  }
+}
+
+resource "aws_api_gateway_stage" "api" {
+  deployment_id = aws_api_gateway_deployment.api.id
+  rest_api_id   = aws_api_gateway_rest_api.api.id
+  stage_name    = "v1"
+}
```
patheard, line 74: You'll also need to add the custom API gateway domain to this, which will need a block like so:
https://github.com/cds-snc/covid-alert-metrics-terraform/blob/main/aws/api_gateway/metrics_gateway.tf#L13-L27

You may be able to use the existing Notify staging ACM certificate for this - it depends on how it's setup.  If it's a wildcard cert for `*.notification.staging.cdssandbox.xyz` you can use it.

aws/common/lambda.tf
```diff
@@ -77,6 +77,65 @@ resource "aws_lambda_function" "ses_receiving_emails" {
   }
 }
 
+resource "aws_lambda_function" "api" {
+  function_name = "api"
+
+  package_type = "Image"
+  image_uri    = "${aws_ecr_repository.api.repository_url}:latest"
```
patheard, line 84: Note that this will fail the first time you try to `terraform apply` since the ECR registry won't exist yet.  Just one of the sequencing gotchas around working with Terraform.

Nothing you need to change though - just one of those "run it again" situations.

aws/common/api_gateway.tf
```diff
@@ -0,0 +1,74 @@
...
+    "application/json" = ""
+  }
+}
+
+resource "aws_api_gateway_deployment" "api" {
+  rest_api_id = aws_api_gateway_rest_api.api.id
+
+  lifecycle {
+    create_before_destroy = true
+  }
```
patheard, line 67: Check the `triggers` block on this example:
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/api_gateway_deployment#terraform-resources

This gives you an implicit dependency on all those resources, which will make sure that:
1. If any change, the API is redeployed; and
2. The API deployment resource won't be created until the listed trigger resources are created.

```hcl
resource "aws_api_gateway_deployment" "example" {
  rest_api_id = aws_api_gateway_rest_api.example.id

  triggers = {
    # NOTE: The configuration below will satisfy ordering considerations,
    #       but not pick up all future REST API changes. More advanced patterns
    #       are possible, such as using the filesha1() function against the
    #       Terraform configuration file(s) or removing the .id references to
    #       calculate a hash against whole resources. Be aware that using whole
    #       resources will show a difference after the initial implementation.
    #       It will stabilize to only change when resources change afterwards.
    redeployment = sha1(jsonencode([
      aws_api_gateway_resource.example.id,
      aws_api_gateway_method.example.id,
      aws_api_gateway_integration.example.id,
    ]))
  }

  lifecycle {
    create_before_destroy = true
  }
}

```

This is probably something we should add to our scan-websites code 😉

aws/common/api_gateway.tf
```diff
@@ -0,0 +1,74 @@
...
+}
+
+resource "aws_api_gateway_stage" "api" {
```
patheard, line 70: It'd be helpful to create an `access_log` block on this resource, similar to what I've [done on this project](https://github.com/patheard/aws-fastapi-lambda/blob/main/infra/modules/api-gateway/api-gateway.tf#L40-L43).

It'll also clear up the other `tfsec` warning you're getting on the PR.

aws/common/lambda.tf
```diff
@@ -77,6 +77,65 @@ resource "aws_lambda_function" "ses_receiving_emails" {
...
+
+  role    = aws_iam_role.api.arn
```
patheard, line 86: I'm not seeing this role?  You'll need to define a role that can be assumed by the Lambda function and lets it interact with the VPC.  Here's an [example of how I've done it](https://github.com/patheard/aws-fastapi-lambda/blob/main/infra/modules/api-gateway/lambda.tf#L35-L43).

patheard (commented):
Alrighty, had a look and I've got some comments/suggestions.  Happy to chat if you want to go over anything!

.github/workflows/merge_to_main_staging.yml
```diff
@@ -96,6 +118,11 @@ jobs:
           cd env/staging/cloudfront
           ../../../bin/terragrunt apply --terragrunt-non-interactive -auto-approve
 
+      - name: Apply aws/lambda-api
+        run: |
+          cd env/staging/lambda-api
+          ../../../bin/terragrunt apply --terragrunt-non-interactive -auto-approve
+
```
patheard, line 125: Can you add this same job step to [.github/workflows/pull_requests.yml](https://github.com/cds-snc/notification-terraform/blob/feature/terraform-for-lambda-api/.github/workflows/pull_requests.yml).  That'll let us see the plan changes for the new Terraform.

aws/lambda-api/iam.tf
```diff
@@ -0,0 +1,115 @@
...
+}
+
+data "aws_iam_policy_document" "api_policies" {
+
+  statement {
+
+    effect = "Allow"
+
+    actions = [
+      "logs:CreateLogGroup",
```
patheard, line 35: You don't need this since the Terraform already creates the log group.  

It doesn't hurt to have it, but it can lead to weird situations if you ever destroy the infrastructure, buffered logs re-create the log group unexpectedly, which causes the next `terraform apply` to fail with a "Log group already exists" error.

aws/lambda-api/iam.tf
```diff
@@ -0,0 +1,115 @@
...
+      "logs:CreateLogStream",
+      "logs:PutLogEvents"
+    ]
+
+    resources = [
+      "*"
```
patheard, line 41: This works, but for better security, you can scope it down the target log group ARN:
```hcl
aws_cloudwatch_log_group.api_gateway_log_group.arn
```

aws/lambda-api/iam.tf
```diff
@@ -0,0 +1,115 @@
...
+        "sqs:*",
+        "sns:Publish",
+        "securityhub:BatchImportFindings",
+        "s3:*"
+      ],
+      "Effect": "Allow",
+      "Resource": "*"
+    }
+  ]
+}
```
patheard, line 108: I'd recommend sticking with the `data "aws_iam_policy_document" "policy_name"` approach you've used above just for consistency.  

There's nothing wrong, Terraform-wise, with this approach though.

aws/lambda-api/lambda.tf
```diff
@@ -0,0 +1,75 @@
...
+  timeout = 60
+
+  memory_size = 1024
+  publish = true
+
+  vpc_config {
+    security_group_ids = [
+    var.eks_cluster_securitygroup,
+  ]
+    subnet_ids = var.vpc_private_subnets
```
patheard, line 17: nit: I'm not sure if your pull request workflow has a Terraform format check, but this would fail if yes (it would still run fine though).  

An easy way to format everything up nicely is to run this at the top level of your `*.tf` folder:
```sh
terraform fmt -recursive
```

aws/lambda-api/lambda.tf
```diff
@@ -0,0 +1,75 @@
...
+  source_arn    = "${aws_api_gateway_rest_api.api.execution_arn}/*/*"
+}
+
+resource "aws_lambda_permission" "api_2" {
+  statement_id  = "AllowAPIGatewayInvoke2"
+  action        = "lambda:InvokeFunction"
+  function_name = aws_lambda_function.api.function_name
+  principal     = "apigateway.amazonaws.com"
+  source_arn    = "${aws_api_gateway_rest_api.api.execution_arn}/*/*/*"
+}
```
patheard, line 69: Why the two `aws_lambda_permission` resources?

aws/lambda-api/variables.tf
```diff
@@ -0,0 +1,94 @@
+variable "admin_client_secret" {
+  type = string
+  sensitive = true
```
patheard, line 3: nit: same comment about `terraform fmt`

env/staging/lambda-api/terragrunt.hcl
```diff
@@ -0,0 +1,44 @@
...
+
+  # Configure mock outputs for the `validate` command that are returned when there are no outputs available (e.g the
+  # module hasn't been applied yet.
+  mock_outputs_allowed_terraform_commands = ["validate"]
+  mock_outputs = {
+    kms_arn = ""
+    vpc_private_subnets = [
+      "subnet-001e585d12cce4d1e",
+      "subnet-08de34a9e1a7458dc",
+      "subnet-0af8b8402f1d605ff",
```
patheard, line 16: You can also set these as blank strings in the mock output and it'll still work.  There's no issue having a non-empty mock value specified though.

env/staging/lambda-api/terragrunt.hcl
```diff
@@ -0,0 +1,44 @@
+dependencies {
+  paths = ["../common", "../eks"]
+}
+
+dependency "common" {
+  config_path = "../common"
+
+  # Configure mock outputs for the `validate` command that are returned when there are no outputs available (e.g the
+  # module hasn't been applied yet.
+  mock_outputs_allowed_terraform_commands = ["validate"]
```
patheard, line 10: I usually set this to the full list of TF commands that might need mocked output:
```hcl
```

env/staging/lambda-api/terragrunt.hcl
```diff
@@ -0,0 +1,44 @@
...
+  }
+}
+
+include {
+  path = find_in_parent_folders()
+}
+
+inputs = {
+  eks_cluster_securitygroup = dependency.eks.outputs.eks-cluster-securitygroup
+  vpc_private_subnets       = dependency.common.outputs.vpc_private_subnets
```
patheard, line 39: There's no harm adding your non-secret variable values here - I find it makes it easier to maintain, but it's a preference thing - always using `TF_VAR_variable_name` is totally legit.
```hcl
aws_pinpoint_region = "us-west-2"
redis_enabled       = "1"
```

patheard (changes requested):
Looking good Steve!  I've added more review comments - if anything's not clear, let me know and we can chat about it.

.github/workflows/merge_to_main_staging.yml
sastels: Added to the end:
```
      - name: Plan aws/lambda-api
        run: |
          cd env/${{env.TARGET_ENV_PATH}}/lambda-api
          ../../../bin/terragrunt plan --terragrunt-non-interactive
```

aws/lambda-api/iam.tf
sastels: @patheard I moved it into the `data "aws_iam_policy_document" "api_policies"` block above - does this make sense, or would it be better to have its own block?

aws/lambda-api/lambda.tf
sastels: 🤷‍♂️  Steve M works in mysterious ways.
Smoke tests all run without it though. I think it might have something to do with concurrency.

aws/lambda-api/lambda.tf
```diff
@@ -0,0 +1,67 @@
+resource "aws_lambda_function" "api" {
+  function_name = "api-${var.env}"
+
+  package_type = "Image"
+  image_uri    = "${aws_ecr_repository.api.repository_url}:latest"
```
patheard, line 5: One minor comment - when you move this to prod, you'll want to switch away from `latest` to a defined image tag.  Usually I do it as a variable in the `terragrunt.hcl` but you can also do it with a manifest style approach.

aws/lambda-api/iam.tf
patheard: Yup, totally makes sense in the same block.  

There's actually a limit to the number of policies that can be attached to a role (10 I think) and so having multiple `statement` elements in a single policy is a way to work around this limit.

aws/lambda-api/lambda.tf
patheard: `¯\_(ツ)_/¯`

I'd vote for removing it and seeing if anything breaks.

.github/workflows/pull_requests.yml
```diff
@@ -104,3 +104,8 @@ jobs:
         run: |
           cd env/${{env.TARGET_ENV_PATH}}/cloudfront
           ../../../bin/terragrunt plan --terragrunt-non-interactive
+
+      - name: Plan aws/lambda-api
+        run: |
+          cd env/${{env.TARGET_ENV_PATH}}/lambda-api
+          ../../../bin/terragrunt plan --terragrunt-non-interactive
```
patheard, line 111: Something you could try (certainly not required) is to switch to the [cds-snc/terraform-plan](https://github.com/cds-snc/terraform-plan) GitHub action.  This will run `terraform plan` and output the plan as a comment on the PR.  

To use it, you'd just need to change the GitHub workflow like so:
1. [Install Terragrunt globally](https://github.com/cds-snc/platform-mvp-ircc/blob/e23f953ec1414cd3755bdf1665fbe295defef664/.github/workflows/terragrunt-plan-prod.yml#L52-L57) in the workflow.
2. [Use the action for each folder](https://github.com/cds-snc/platform-mvp-ircc/blob/e23f953ec1414cd3755bdf1665fbe295defef664/.github/workflows/terragrunt-plan-prod.yml#L91-L99) (the `if` piece isn't required).

patheard (approved):
Looks good!  I had a few non-blocking comments, but I'd say you're good to merge!

When you're back from vacation, lets talk next steps again.

.github/workflows/pull_requests.yml
sastels: Thinking a bit more about it, I think we can do this in another PR.

aws/lambda-api/api_gateway.tf
```diff
@@ -0,0 +1,127 @@
...
+  access_log_settings {
+    destination_arn = aws_cloudwatch_log_group.api_gateway_log_group.arn
+    format          = "{\"requestId\":\"$context.requestId\", \"ip\": \"$context.identity.sourceIp\", \"caller\":\"$context.identity.caller\", \"requestTime\":\"$context.requestTime\", \"httpMethod\":\"$context.httpMethod\", \"resourcePath\":\"$context.resourcePath\", \"status\":\"$context.status\", \"responseLength\":\"$context.responseLength\"}"
+  }
+}
+
+resource "aws_cloudwatch_log_group" "api_gateway_log_group" {
+  name = "api_gateway_log_group"
+  tags = {
+    Environment = "staging"
```
jimleroyer, line 108: Should we replace the environment hard coded value with the variable? or it's unrelated to the environment in which Notify runs?

env/staging/lambda-api/terragrunt.hcl
```diff
@@ -0,0 +1,49 @@
...
+  aws_pinpoint_region       = "us-west-2"
+  redis_enabled             = "1"
+  sqlalchemy_pool_size      = "256"
+  api_image_tag             = "latest"
```
jimleroyer, line 43: I assume we'll need to come up with a release strategy that will write a value with a specific SHA ID to specify the image tag, is this right? I wonder because in case we want to revert and not use the latest available image in the image repository.

aws/lambda-api/lambda.tf
```diff
@@ -0,0 +1,67 @@
...
+      DOCUMENT_DOWNLOAD_API_HOST     = var.document_download_api_host
+    }
+  }
+
+  lifecycle {
+    ignore_changes = [
+      image_uri,
+    ]
+  }
+}
```
jimleroyer, line 53: Beautiful

jimleroyer (commented):
Excellent work Steves~! 🙌

aws/lambda-api/api_gateway.tf
sastels: Hrm, yeah, there’s a variable `var.env` that we should probably use here.

env/staging/lambda-api/terragrunt.hcl
sastels: This is the staging config so this is probably good. Reverting api would change latest to point to the new reverted version. The prod version of this file should be different. And somehow use the manifest repo?

env/staging/lambda-api/.terraform.lock.hcl
```diff
@@ -0,0 +1,21 @@
...
+    "zh:0d4af59266668089790f5a7bdeb25642ba750fb5dc7934fe28d1cc36310ba495",
+    "zh:1413ff4d445678c096d46e8957e27320df94561354955d7bc5d8054b6df7b299",
+    "zh:19d614259f7ce16b50ec07868404b58749702baaf86bcd14fbaea2756e1c9f25",
+    "zh:2d148ff632da25852622b06b5be9f5a0b6d509621a002a47338f96509021945b",
+    "zh:3959a1d989c99f3e7cdd5de07eb3e7df7a85e19677488278c77ab753dd7127e6",
+    "zh:5d8d65b458a8934dc67d22904da368b5bc3a77fb9c900ac89c54e736a221b76f",
+    "zh:94d5660e56118fcaa40fccaff960a9bf4166b7b0e7fedeb21b2402c8fc7b4cb1",
+    "zh:a6002ecc23ebc468ccac6f36c0ed7cc95de3223ef6b100e6c81762d22cc14077",
+    "zh:b0880c82bc2ad395ef3dbf5a592a23e65bf943df8995d5d4238740f96a02f529",
+    "zh:ee65f3d2c13653e0828222a63fb832f98e9835b84443eeca00ce36ae39783c08",
```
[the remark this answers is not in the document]
sastels: This is a good point… @patheard should this be here? I thought I read somewhere that our state was stored  in AWS?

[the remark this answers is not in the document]
patheard: Generally yes, but your project is bit odd in this.  The `.terraform.lock.hcl` locks the versions of the providers/modules that are installed by the `terraform init` command (like the `package-lock.json` file in a Node.js project).  

The first time `terraform init` runs, it checks the version constraints specified for the providers/modules and installs the ones that match.  It then creates `.terraform.lock.hcl` to specify exactly which version of each provider/module was installed.

The next time you run `terraform init`, instead of using the version constraints, it installs the exact version from `.terraform.lock.hcl`.  In other words, it pins versions to have consistent installs and `terraform plan/apply` runs.

Your project doesn't do this because the `terraform init` command is always run in GitHub Actions, so the `.terraform.lock.hcl` file gets generated, but never committed to the repo.  As a result, there's a small chance that at some point in the future you could end up with a broken `terraform plan/apply` run because a provider or module upgrades, but still matches your version constraints.

Personally I would say each of your modules should have a `.terraform.lock.hcl` committed.

env/staging/lambda-api/terragrunt.hcl
jimleroyer: Ok get it.. and yeah, ideally if we could leverage the manifest repository or have something similar, it would be great.

PR #282, opened 2021-07-27
Add RDS proxies

jimleroyer:
# Summary | Résumé

Adding RDS proxies via Terraform. This is part of [the ADR's recommendation](https://github.com/cds-snc/notification-adr/pull/1).

# Test instructions | Instructions pour tester la modification

A similar infrastructure than what is proposed in the Terraform has been setup manually in the *staging* environment. Our goal is to replicate that environment's components with the Terraform code and then target the staging Notify stack to the RDS proxy endpoints.

Also, the RDS proxy have been tested with a local Notify instance running on my computer. The RDS endpoint was tunneled via the jumpbox to my local machine. I had to modify some code of the *notification-api* component to make it work but it was successful. This was done with the RDS manually created component. I will need to replicate the same with the infrastructure automatically created with Terraform once pushed.

# List of changes the Terraform plan will bring in

Compare the [manually created RDS proxy and its artifacts](https://ca-central-1.console.aws.amazon.com/rds/home?region=ca-central-1#proxy:id=notification-canada-ca-staging-rds-proxy-writer) with the list of changes that this Terraform plan will apply once merged:

```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create
  ~ update in-place
 <= read (data resources)

Terraform will perform the following actions:

  # aws_rds_cluster.notification-canada-ca will be updated in-place
  ~ resource "aws_rds_cluster" "notification-canada-ca" {
        id                                  = "notification-canada-ca-staging-cluster"
      ~ master_password                     = (sensitive value)
        tags                                = {
            "CostCenter" = "notification-canada-ca-staging"
        }
        # (32 unchanged attributes hidden)
    }

  # aws_secretsmanager_secret.superuser will be created
  + resource "aws_secretsmanager_secret" "superuser" {
      + arn                     = (known after apply)
      + description             = "Database superuser postgres, database connection values"
      + id                      = (known after apply)
      + name                    = "postgres"
      + name_prefix             = (known after apply)
      + policy                  = (known after apply)
      + recovery_window_in_days = 30
      + rotation_enabled        = (known after apply)
      + rotation_lambda_arn     = (known after apply)
      + tags                    = {
          + "CostCenter" = "notification-canada-ca-staging"
        }
      + tags_all                = {
          + "CostCenter" = "notification-canada-ca-staging"
        }

      + rotation_rules {
          + automatically_after_days = (known after apply)
        }
    }

  # aws_secretsmanager_secret_version.superuser will be created
[...]

PR #309, opened 2021-10-29
add new separate terraform plans for staging and production

sastels:
# Summary | Résumé

Separate the terraform plan for staging and production. Use the new action for the plan to better surface results.

Note that changes to the terraform / terragrunt files were to fix formatting issues surfaced by `terraform fmt`
(with the exception of the variable removed in `aws/rds/variables.tf`

      langues officielles?
      fonctionnalité existante?


aws/rds/variables.tf
```diff
@@ -6,13 +6,9 @@ variable "kms_arn" {
   type = string
 }
 
-variable "rds_server_db_user" {
-  type = string
-}
-
```
sastels, line 12: This variable is not used in the code. It was added in https://github.com/cds-snc/notification-terraform/pull/282 where it was briefly used, but in the end the code that used it was deleted from the PR in https://github.com/cds-snc/notification-terraform/commit/8b36b21bee50aa0c873084bc7fa399391d2cbe47 (it was in the `secrets.tf` file that was deleted in that commit)

sastels:
> ## Production: rds
> **❌   Terraform Format:** `failed` **❌   Terraform Plan:** `failed`
> 
> **⚠️   Format:** run `terraform fmt` to fix the following:
> 
> ```shell
> rds.tf
> rds_proxy.tf
> variables.tf
> ```

Note that this is failing for production (but not staging) because the production `terragrunt plan` is using the version of code specified by `INFRASTRUCTURE_VERSION.txt`, so does not have the formatting fixes contained in this PR.

sastels:
huh maybe I shouldn't be running `terragrunt plan` in `common`, @patheard ?

.github/workflows/terragrunt_plan_production.yml
```diff
@@ -0,0 +1,165 @@
+name: "Terragrunt plan PRODUCTION"
+
+on:
+  pull_request:
+    paths:
+      - "aws/**"
+      - "env/production/**"
+      - "env/terragrunt.hcl"
```
patheard, line 8: To cut down on PR noise, it might make sense to set this to the same triggers as the `merge_to_main_production.yml` paths:
https://github.com/cds-snc/notification-terraform/blob/main/.github/workflows/merge_to_main_production.yml#L8-L10

That way you only see Prod plan comments if you're expecting to have `terraform apply` run against Prod when you merge.

patheard (approved):
Looks really good Steve!  Just had a minor comment for the production plan `path` triggers, but it's a preference thing.

patheard:
> huh maybe I shouldn't be running `terragrunt plan` in `common`, @patheard ?

Based on your infra, I think it makes sense to run `terragrunt plan` in `common`.  The recreation of that Slack Lambda `null_resource` is unfortunately buried in the module you're using, so it's not something we can ignore.

That being said, we have our own Notify Slack module without this behaviour that might do what you need:
https://github.com/cds-snc/terraform-modules/tree/main/notify_slack

aws/rds/variables.tf
jimleroyer: Thanks for cleaning it up.

jimleroyer (approved):
:shipit: (on Monday!)

jimleroyer:
Do we know why Terraform still complain about the Slack modules in his plans?

CalvinRodo:
> Do we know why Terraform still complain about the Slack modules in his plans?

The timestamp trigger is forcing that resource to get tainted.

jimleroyer:
@CalvinRodo Seems like this is linked to the Slack aws notify module and has been addressed:
https://github.com/terraform-aws-modules/terraform-aws-notify-slack/issues/84

I opened a PR to disable automatic overwriting of the Slack lambda:
https://github.com/cds-snc/notification-terraform/pull/310

sastels:
It's still complaining that it's... maybe not going to do anything? This might be as good as it gets unless we replace with SRE's slack integration code.

PR #312, opened 2021-11-01
HOT-FIX: Replace timestamp() to "2012-10-17"

iokpala:
# Summary | Résumé

This HOT-FIX seeks to replace timestamp() with "2012-10-17", this
should fix this error [1]

[1] https://github.com/cds-snc/notification-terraform/runs/4071306016?check_suite_focus=true#step:15:80


sastels (approved):
🚀

PR #317, opened 2021-11-08
Update ENVs and define S3 bucket with permissions 

iokpala:
Related to https://github.com/cds-snc/notification-api/pull/1395

# Summary | Résumé

To publish performance test results to s3 (for storage), this commit defines an s3 terraform script that reuses the SRE team's S3 terraform module [1].

Within is also defined aws_iam_policy that allow for CRUD activities [2],

Also within we have updated the environment variables and 

[1] github.com/cds-snc/terraform-modules?ref=v0.0.38//S3
[2] https://registry.terraform.io/modules/cloudposse/s3-bucket/aws/latest/examples/complete?tab=inputs


env/staging/performance-test/terragrunt.hcl
```diff
@@ -39,7 +39,7 @@ inputs = {
   api_image_tag               = "latest"
   eks_cluster_securitygroup   = dependency.eks.outputs.eks-cluster-securitygroup
   vpc_private_subnets         = dependency.common.outputs.vpc_private_subnets
-  aws_pinpoint_region         = "us-west-2"
+  aws_pinpoint_region         = "ca-central-1"
```
jimleroyer, line 42: Good catch

aws/performance-test/lambda.tf
```diff
@@ -15,6 +15,26 @@ resource "aws_lambda_function" "performance-test" {
...
+      LOAD_TEST_PHONE_NUMBER                      = var.load_test_phone_number 
+      LOAD_TEST_EMAIL                             = var.load_test_email
+      LOAD_TEST_AWS_S3_BUCKET                     = var.load_test_aws_s3_bucket
+      LOAD_TEST_CSV_DIRECTORY_PATH                = var.load_test_csv_directory_path
+      LOAD_TEST_DOMAIN                            = var.load_test_domain
+      LOAD_TEST_SMS_TEMPLATE_ID                   = var.load_test_sms_template_id
+      LOAD_TEST_BULK_EMAIL_TEMPLATE_ID            = var.load_test_bulk_email_template_id
+      LOAD_TEST_EMAIL_TEMPLATE_ID                 = var.load_test_email_template_id
+      LOAD_TEST_EMAIL_WITH_ATTACHMENT_TEMPLATE_ID = var.load_test_email_with_attachment_template_id
+      LOAD_TEST_EMAIL_WITH_LINK_TEMPLATE_ID       = var.load_test_email_with_link_template_id
```
jimleroyer, line 34: Can we keep the same semantic around performance tests and rename these variables `PERF_TEST...` instead of `LOAD_TEST...`?

env/staging/performance-test/terragrunt.hcl
iokpala: Cheers

aws/performance-test/lambda.tf
```diff
@@ -15,6 +15,26 @@ resource "aws_lambda_function" "performance-test" {
       image_uri,
     ]
   }
+
+  environment {
+    variables = {
+      AWS_ACCESS_KEY_ID                           = var.aws_access_key_id
+      AWS_SECRET_ACCESS_KEY                       = var.aws_secret_access_key
```
patheard, line 22: You should be able to get rid of these variables since you're using an IAM policy to grant access to the S3 bucket from the Lambda.

aws/performance-test/variables.tf
```diff
@@ -2,3 +2,67 @@ variable "schedule_expression" {
   default     = "cron(0 0 * * ? *)"
   description = "This aws cloudwatch event rule scheule expression that specifies when the scheduler runs."
 }
+variable "env" {
+  default     = "staging"
+  description = "This helps identify the environment within which would run."
+}
```
patheard, line 8: You can get rid of this too since the common `env` variable should have the value you need.

aws/performance-test/variables.tf
```diff
@@ -2,3 +2,67 @@ variable "schedule_expression" {
...
+variable "aws_access_key_id" {
+  default     = ""
+  description = "This helps identify aws access key."
+}
+variable "aws_secret_access_key" {
+  default     = ""
+  description = "This helps identify aws access key."
+}
```
patheard, line 16: These can go as well - we'll rely on the IAM policy instead so you don't have to manage a new IAM user with an access key and secret.

aws/performance-test/variables.tf
```diff
@@ -2,3 +2,67 @@ variable "schedule_expression" {
...
+variable "aws_default_output" {
+  default     = "json"
+  description = "This helps identify aws access key."
```
patheard, line 19: nit: could you make the variable descriptions unique.

patheard (changes requested):
Looks like a good start!  Had some comments around how to get the Terraform plan working and how we can switch to just using the IAM policy change instead of AWS access and secret keys.

aws/performance-test/variables.tf
```diff
@@ -2,3 +2,67 @@ variable "schedule_expression" {
...
+}
+variable "aws_access_key_id" {
+  default     = ""
+  description = "This helps identify aws access key."
+}
+variable "aws_secret_access_key" {
+  default     = ""
+  description = "This helps identify aws access key."
+}
+variable "aws_default_output" {
```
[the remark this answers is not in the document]
iokpala: It refers to the putput format example text, json etc see https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-output-format.html

aws/performance-test/variables.tf
```diff
@@ -2,3 +2,67 @@ variable "schedule_expression" {
...
+  default     = "ca-central-1"
+  description = "This helps identify aws access key."
+}
+variable "load_test_phone_number" {
+  default     = "16132532222"
+  description = "This helps identify aws access key."
+}
+variable "load_test_email" {
+  default     = "success@simulator.amazonses.com"
+  description = "This helps identify aws access key."
```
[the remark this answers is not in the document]
iokpala: Will correct that

aws/performance-test/lambda.tf
iokpala: Yeah, well this is a continuation of the definition from the api repo.. See https://github.com/cds-snc/notification-api/blob/master/tests-perf/locust/locust-notifications.py#L36-L43 and https://github.com/cds-snc/notification-api/blob/89e6d832443d337c88b27f5a38a0e2eb12413bc3/tests-perf/ops/Dockerfile#L22-L30

aws/performance-test/variables.tf
iokpala: Will correct it

aws/performance-test/variables.tf
iokpala: Absolutely

aws/performance-test/variables.tf
iokpala: I defined this after getting an error that it was not declated.. Will take a look at common

aws/performance-test/lambda.tf
iokpala: Will remove

aws/performance-test/variables.tf
```diff
@@ -1,4 +1,60 @@
...
+  description = "This helps identify delivery phone number."
+}
+variable "load_test_email" {
+  type        = string
+  default     = "success@simulator.amazonses.com"
+  description = "This helps identify delivery email address."
+}
+variable "load_test_aws_s3_bucket" {
+  type        = string
+  default     = "notify-performance-test-results-${var.env}"
```
patheard, line 18: This is the line that's giving you grief in your `terraform plan`.  Unfortunately you can't reference variables within a `variable` block.

aws/performance-test/lambda.tf
jimleroyer: Yeah I see, so we are misaligned in there as well. The directory call these performance tests. The image in this PR calls these performance test. I would like to aim for one semantic to avoid confusion on what is what. What is your suggestion?

aws/performance-test/variables.tf
jimleroyer: So we'll have to concat the strings where it's used, right?

aws/performance-test/variables.tf
```diff
@@ -1,4 +1,60 @@
...
+}
+variable "load_test_csv_directory_path" {
+  type        = string
+  default     = "/tmp/notify_performance_test"
+  description = "This helps identify the csv directory."
+}
+variable "load_test_domain" {
+  type        = string
+  default     = "https://api.staging.notification.cdssandbox.xyz"
+  description = "This helps identify performance test domain."
```
jimleroyer, line 29: nitpick: `Identifies performance test domain` or `Performance test domain` might be better worded, as well for the other ones.

aws/performance-test/variables.tf
jimleroyer: Thanks for the explanation. This was the TF default I guess? as I see we dropped this variable?

aws/performance-test/variables.tf
patheard: It just means you can't use the variable to set a `default` value here.  So yup, string concat if you're doing this in a `terragrunt.hcl` file, or just like so if you're using it in a `*.tf` file:
https://github.com/cds-snc/notification-terraform/blob/feature/publish-performance-test-results/aws/performance-test/s3.tf#L3

iokpala: Yikes

iokpala: I have xtried a few things and nothing really worked.. so will just go with a static value

aws/performance-test/s3.tf
```diff
@@ -0,0 +1,31 @@
...
+    actions = [
+      "s3:DeleteObject",
+      "s3:GetObject",
+      "s3:PutObject",
+      "s3:ListBucket"
+    ]
+
+    resources = [
+      aws_s3_bucket.reliability_file_storage.arn,
+      "${aws_s3_bucket.reliability_file_storage.arn}/*"
```
patheard, line 24: Missed this the last time round - this will need to be updated to the ARN of the performance bucket like so:
```hcl
module.notify_performance_test_results.s3_bucket_arn,
"${module.notify_performance_test_results.s3_bucket_arn}/*"
```

iokpala: Thanks will update it

aws/performance-test/s3.tf
```diff
@@ -1,7 +1,7 @@
 module "notify_performance_test_results" {
   source            = "github.com/cds-snc/terraform-modules?ref=v0.0.38//S3"
-  bucket_name       = "notify-performance-test-results-${var.env}"
-  billing_tag_value = "notification-canada-ca-${var.env}"
```
jimleroyer: Can we not keep this variable?

aws/performance-test/variables.tf
```diff
@@ -15,7 +15,7 @@ variable "load_test_email" {
 }
 variable "load_test_aws_s3_bucket" {
   type        = string
-  default     = "notify-performance-test-results-${var.env}"
+  default     = "notify-performance-test-results-staging"
```
jimleroyer, line 18: We'd remove the env variable (and static reference) from this and concat it where it is used.

aws/performance-test/variables.tf
jimleroyer: I don't think that @patheard meant to go full static values, but rather to use the variables in places to concat these. I can show you in places where we do it in the TF code if you want.

aws/performance-test/variables.tf
iokpala: @patheard  Would it be possible to concat this from the terraform modules??

aws/performance-test/s3.tf
```diff
@@ -0,0 +1,31 @@
+module "notify_performance_test_results" {
+  source            = "github.com/cds-snc/terraform-modules?ref=v0.0.38//S3"
+  bucket_name       = "notify-performance-test-results-staging"
```
patheard, line 3: These can become:
```hcl
  bucket_name       = "notify-performance-test-results-${var.env}"
  billing_tag_value = "notification-canada-ca-${var.env}"
```

iokpala: Thanks Pat.. I have added it back

aws/performance-test/lambda.tf
jimleroyer: @iokpala You asked for another review round but you have not replied to me on this yet.

iokpala: I see merits in this Would it be Ok to have this request in a new PR?? I ask this because the `LOAD_TEST` prefix is already being used.. My appeal is that this be done in uniformity given previous acceptance

aws/performance-test/variables.tf
iokpala: I beg to differ as this appears deeply preferential and open for interpret within different linguistic context

aws/performance-test/variables.tf
iokpala: With regards to this.. I misplaced the variables and had a session with Pat, where he pointed me in the right direction.. I have since reverted this change back.

aws/performance-test/s3.tf
iokpala: Yes this change has been reverted

aws/performance-test/lambda.tf
jimleroyer: Alright

aws/performance-test/variables.tf
jimleroyer: Yeah it might be preferential, I try to favor short and concise message/doc. The interpretation and context is quite limited to Terraform and environment variable handling here though.

aws/performance-test/variables.tf
jimleroyer: Can we just try it?

aws/performance-test/variables.tf
iokpala: This change has been made to reflect "Identifies"

aws/performance-test/variables.tf
jimleroyer: oh wrong thread, sorry

jimleroyer (approved):

patheard (approved):
🎉

PR #318, opened 2021-11-10
COSMETIC: Change LOAD_TEST to PERF_TEST

iokpala:
Related to https://github.com/cds-snc/notification-terraform/pull/317

# Summary | Résumé

This is makes an update from LOAD_TEST to PERF_TEST.

### DIFF
--- a/aws/performance-test/lambda.tf
+++ b/aws/performance-test/lambda.tf
@@ -18,16 +18,16 @@ resource "aws_lambda_function" "performance-test" {
 
   environment {
     variables = {
-      LOAD_TEST_PHONE_NUMBER                      = var.load_test_phone_number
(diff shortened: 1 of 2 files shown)


jimleroyer:
I think if that PR targets the `feature/publish-performance-test-results` branch, the #317 will basically squash the commit. I mention it as you mentioned to have a new PR to preserve the commit history (but that might not work as setup right now -- and there would be no difference with this PR and applying the fix directly in #317). 

I don't mind, I can approve and merge it.

iokpala:
Cheers Jimmy

jimleroyer:
So I will assume this is a yes to merge it. 😉

jimleroyer (approved):

PR #321, opened 2021-11-15
Reinstated the fix to only support TLS 2.0 connections

jimleroyer:
# Summary | Résumé

Related previous changes:

* https://github.com/cds-snc/notification-terraform/pull/246
* https://github.com/cds-snc/notification-terraform/pull/253
* https://github.com/cds-snc/notification-terraform/pull/262

I kept a section of the old addition over there:
https://github.com/cds-snc/notification-terraform/blob/main/aws/eks/alb.tf#L68-L85

It seems to return a fixed unauthorized response. Antoine said in one of the PR that it should be kept. But I am unsure how the fixed response was overriden, possibly that the redirection ALB rule overrode it. Hence I need some SRE 👀 to look if it's okay to keep the fixed response.

We'll deploy in staging and test againt the `4444` API port to see what's the returned response.

      langues officielles?
      fonctionnalité existante?

### DIFF
--- a/aws/eks/alb.tf
+++ b/aws/eks/alb.tf
@@ -32,8 +32,7 @@ resource "aws_alb_listener" "notification-canada-ca" {
   certificate_arn   = var.aws_acm_notification_canada_ca_arn
   # See https://docs.aws.amazon.com/elasticloadbalancing/latest/network/create-tls-listener.html#describe-ssl-policies
   # And https://cyber.gc.ca/en/guidance/guidance-securely-configuring-network-protocols-itsp40062
-  #tfsec:ignore:AWS010 Outdated SSL policy
-  ssl_policy = "ELBSecurityPolicy-2016-08"
+  ssl_policy = "ELBSecurityPolicy-FS-1-2-Res-2019-08"
 
   default_action {
     type             = "forward"
@@ -223,22 +222,6 @@ resource "aws_lb_listener_rule" "api-host-route" {
   condition {
     host_header {
       values = ["api.*"]
-    }
-  }
-}
-
-resource "aws_lb_listener_rule" "api-host-route-legacy-tls" {
-  listener_arn = aws_lb_listener.notification-canada-ca-legacy-tls.arn
-  priority     = 1
-
-  action {
-    type             = "forward"
-    target_group_arn = aws_alb_target_group.notification-canada-ca-api.arn
-  }
-
-  condition {
-    host_header {
-      values = ["api.${var.domain}"]
     }
   }
 }
--- a/aws/eks/alb.tf
+++ b/aws/eks/alb.tf
@@ -77,8 +77,8 @@ resource "aws_lb_listener" "notification-canada-ca-legacy-tls" {
 
     fixed_response {
       content_type = "text/plain"
-      message_body = "Unauthorized"
-      status_code  = "401"
+      message_body = "Forbidden"
+      status_code  = "403"
     }
   }
 }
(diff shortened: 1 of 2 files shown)


aws/eks/securitygroups.tf
```diff
@@ -21,13 +21,6 @@ resource "aws_security_group" "notification-canada-ca-alb" {
     cidr_blocks = ["0.0.0.0/0"] #tfsec:ignore:AWS008
   }
 
-  ingress {
-    protocol    = "tcp"
-    from_port   = 4444
-    to_port     = 4444
-    cidr_blocks = ["0.0.0.0/0"] #tfsec:ignore:AWS008
-  }
```
jimleroyer, line 29: I figured that we might want to remove 4444 redirection if the ALB will return a fixed response.

aws/eks/alb.tf
```diff
@@ -227,22 +226,6 @@ resource "aws_lb_listener_rule" "api-host-route" {
...
-    type             = "forward"
-    target_group_arn = aws_alb_target_group.notification-canada-ca-api.arn
-  }
-
-  condition {
-    host_header {
-      values = ["api.${var.domain}"]
-    }
-  }
-}
```
jimleroyer, line 244: We probably want to remove the forward default action for the legacy TLS.

patheard: Since legacy TLS support is being removed, would it make sense to remove [`aws_lb_listener.notification-canada-ca-legacy-tls`](https://github.com/cds-snc/notification-terraform/blob/9ab47ca25530f89ef19caaeaf3fd2eb5478ab191/aws/eks/alb.tf#L67-L84) as well?

patheard (commented):
Had a question about keeping the legacy TLS listener as well.

aws/eks/alb.tf
jimleroyer: Yeah this is what I am wondering. I think Antoine wanted to keep that one, potentially to indicate that this is not supported anymore to users still using the `4444` port with the fixed response, once the legacy TLS and forward rule have been removed. Would that make sense?

patheard: Ah, gotcha - that makes sense.  In that case, perhaps switching the fixed response to `403 Forbidden` instead since based on the behaviour, this is a request that Notify understands, but will not support:

> [10.4.4 403 Forbidden](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.4)
> The server understood the request, but is refusing to fulfill it. Authorization will not help and the request SHOULD NOT be repeated. If the request method was not HEAD and the server wishes to make public why the request has not been fulfilled, it SHOULD describe the reason for the refusal in the entity. If the server does not wish to make this information available to the client, the status code 404 (Not Found) can be used instead.

aws/eks/securitygroups.tf
patheard: I just realized that if you're planning to keep the fixed response from the legacy TLS listener, you'll need to keep this `ingress` rule to allow users from the internet to hit the ALB (and get the fixed response).

aws/eks/alb.tf
jimleroyer: Alright let me change this. Then I'll merge! 🚀

aws/eks/securitygroups.tf
jimleroyer: Just put it back!

patheard (approved):
Looks good!  Let us see what happens 🚀

PR #338, opened 2021-12-03
Further heartbeat keys

jzbahrai:
# Summary | Résumé

Add further heartbeat keys for both staging and production


.github/workflows/merge_to_main_production.yml
```diff
@@ -16,6 +16,8 @@ env:
   AWS_SECRET_ACCESS_KEY: ${{ secrets.PRODUCTION_AWS_SECRET_ACCESS_KEY }}
   AWS_REGION: ca-central-1
   HEARTBEAT_API_KEY: ${{ secrets.PRODUCTION_HEARTBEAT_API_KEY }}
+  HEARTBEAT_BASE_URL: ${{ secrets.PRODUCTION_HEARTBEAT_BASE_URL }}
+  HEARTBEAT_TEMPLATE_ID: ${{ secrets.PRODUCTION_HEARTBEAT_TEMPLATE_ID }}
```
sastels, line 20: These three heatbeat variables are headed for terraform, so we should set them with something like
```
TF_VAR_heartbeat_api_key: ${{ secrets.PRODUCTION_HEARTBEAT_API_KEY }}
```
as we do with the other `TV_VAR`s below.

We should also include these variables in `terragrunt_plan_production.yml`

.github/workflows/merge_to_main_staging.yml
```diff
@@ -17,6 +17,9 @@ env:
   AWS_ACCESS_KEY_ID: ${{ secrets.STAGING_AWS_ACCESS_KEY_ID }}
   AWS_SECRET_ACCESS_KEY: ${{ secrets.STAGING_AWS_SECRET_ACCESS_KEY }}
   AWS_REGION: ca-central-1
+  HEARTBEAT_API_KEY: ${{ secrets.STAGING_HEARTBEAT_API_KEY }}
+  HEARTBEAT_BASE_URL: ${{ secrets.STAGING_HEARTBEAT_BASE_URL }}
+  HEARTBEAT_TEMPLATE_ID: ${{ secrets.STAGING_HEARTBEAT_TEMPLATE_ID }}
```
sastels, line 22: Same comment as above, and including them in `terragrunt_plan_staging.yml`

sastels (changes requested):

.github/workflows/merge_to_main_production.yml
jzbahrai: will do

.github/workflows/merge_to_main_staging.yml
jzbahrai: updated

jzbahrai:
@sastels also updated this: https://github.com/cds-snc/notification-api/pull/1413 to take the new name of TF_VARS

jimleroyer (approved):
LGTM!~ but we can wait for Steve's approval too as he initiated the review.

sastels (approved):
Awesome! 🚀

PR #340, opened 2021-12-06
Add heartbeat lambda

jzbahrai:
# Summary | Résumé

Add the lambda api for heartbeat.

### DIFF
--- a/aws/heartbeat/ecr.tf
+++ b/aws/heartbeat/ecr.tf
@@ -1,6 +1,5 @@
 resource "aws_ecr_repository" "heartbeat" {
   # The :latest tag is used in Staging
-  #tfsec:ignore:AWS078
 
   name                 = "notify/heartbeat"
   image_tag_mutability = "MUTABLE" #tfsec:ignore:AWS078
--- a/aws/heartbeat/ecr.tf
+++ b/aws/heartbeat/ecr.tf
@@ -1,0 +1,11 @@
+resource "aws_ecr_repository" "heartbeat" {
+  # The :latest tag is used in Staging
+  #tfsec:ignore:AWS078
+
+  name                 = "notify/heartbeat"
+  image_tag_mutability = "MUTABLE" #tfsec:ignore:AWS078
+
+  image_scanning_configuration {
+    scan_on_push = true
+  }
+}
--- a/aws/heartbeat/lambda.tf
+++ b/aws/heartbeat/lambda.tf
@@ -1,88 +1,38 @@
-resource "aws_iam_role" "heartbeat" {
-  name = "heartbeat-iam-role"
-
-  # Terraform's "jsonencode" function converts a
-  # Terraform expression result to valid JSON syntax.
-  assume_role_policy = jsonencode({
-    Version = "2012-10-17"
-    Statement = [
-      {
-        Action = "sts:AssumeRole"
-        Effect = "Allow"
-        Sid    = ""
-        Principal = {
-          Service = "lambda.amazonaws.com"
-        }
-      },
-    ]
-  })
-}
-
-resource "aws_lambda_function" "heartbeat" {
-  function_name = "heartbeat"
-
-  package_type = "Image"
-  image_uri    = "${aws_ecr_repository.heartbeat.repository_url}:latest"
-
-  role    = aws_iam_role.heartbeat.arn
-  timeout = 60
-
-  memory_size = 1024
-  publish     = true
-
-  lifecycle {
-    ignore_changes = [
-      image_uri,
-    ]
-  }
+module "heartbeat" {
+  source                   = "github.com/cds-snc/terraform-modules?ref=v0.0.45//lambda"
+  name                     = "heartbeat"
+  billing_tag_value        = var.billing_tag_value
+  ecr_arn                  = aws_ecr_repository.heartbeat.arn
+  enable_lambda_insights   = true
+  image_uri                = "${aws_ecr_repository.heartbeat.repository_url}:latest"
+  timeout                  = 60
+  memory                   = 1024
 
   environment {
     variables = {
-      TF_VAR_heartbeat_api_key                    = var.heartbeat_api_key
-      TF_VAR_heartbeat_base_url                   = var.heartbeat_base_url
-      TF_VAR_heartbeat_template_id                = var.heartbeat_template_id
+      heartbeat_api_key                    = var.heartbeat_api_key
+      heartbeat_base_url                   = var.heartbeat_base_url
+      heartbeat_template_id                = var.heartbeat_template_id
     }
   }
 }
 
 resource "aws_cloudwatch_event_target" "heartbeat" {
-  arn  = aws_lambda_function.heartbeat-test.arn
-  rule = aws_cloudwatch_event_rule.heartbeat-testing.id
-
-  input_transformer {
-    input_paths = {
-      instance = "$.detail.instance",
-      status   = "$.detail.status",
-    }
-    input_template = <<-EOF
-    {
-      "instance_id": <instance>,
-      "instance_status": <status>
-    }
-    EOF
-  }
+  arn  = module.heartbeat.function_arn
+  rule = aws_cloudwatch_event_rule.heartbeat_testing.id
 }
 
-resource "aws_cloudwatch_event_rule" "heartbeat-testing" {
-  name                = "heartbeat-testing"
-  description         = "heartbeat-testing event rule"
+resource "aws_cloudwatch_event_rule" "heartbeat_testing" {
+  name                = "heartbeat_testing"
+  description         = "heartbeat_testing event rule"
   schedule_expression = var.schedule_expression
-  depends_on          = [aws_lambda_function.heartbeat-test]
+  depends_on          = [module.heartbeat]
 }
 
 resource "aws_lambda_permission" "allow_cloudwatch" {
   statement_id  = "AllowExecutionFromCloudWatch"
   action        = "lambda:InvokeFunction"
-  function_name = aws_lambda_function.heartbeat.function_name
+  function_name = module.heartbeat.function_name
   principal     = "events.amazonaws.com"
-  source_arn    = aws_lambda_function.heartbeat.arn
-  qualifier     = aws_lambda_alias.heartbeat-lambda-alias.name
+  source_arn    = module.heartbeat.function_arn
 }
-
-resource "aws_lambda_alias" "heartbeat-lambda-alias" {
-  name             = "heartbeat-lambda-alias"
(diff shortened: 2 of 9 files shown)


aws/heartbeat/lambda.tf
```diff
@@ -0,0 +1,88 @@
...
+      "instance_status": <status>
+    }
+    EOF
+  }
+}
+
+resource "aws_cloudwatch_event_rule" "heartbeat-testing" {
+  name                = "heartbeat-testing"
+  description         = "heartbeat-testing event rule"
+  schedule_expression = var.schedule_expression
```
jzbahrai, line 69: I couldnt figure out where this should be set?

aws/heartbeat/variables.tf
```diff
@@ -0,0 +1,22 @@
...
+}
+
+variable "heartbeat_base_url" {
+  sensitive   = true
+  type        = list(string)
+  description = "Identifies delivery phone number."
+}
+
+variable "heartbeat_template_id" {
+  sensitive   = true
```
jzbahrai, line 14: change sensitive

sastels (changes requested):
You also need to add then new variables as well as a heartbeat section to the github actions:
terragrunt_plan_staging.yml
terragrunt_plan_production.yml
merge_to_main_staging.yml
merge_to_main_production.yml

jzbahrai:
@sastels yup done in this PR: https://github.com/cds-snc/notification-terraform/pull/338
also made the updates we talked about.

env/staging/heartbeat/terragrunt.hcl
```diff
@@ -0,0 +1,13 @@
...
+
+inputs = {
+  billing_tag_key                             = "CostCenter"
+  billing_tag_value                           = "notification-canada-ca-staging"
+  schedule_expression                         = "cron(*/1 * * * *)"
+}
+
+terraform {
+  source = "../../../aws/heartbeat"
+}
```
jimleroyer, line 13: We could add the production values right away as the values are predictable.

aws/heartbeat/ecr.tf
```diff
@@ -0,0 +1,11 @@
+resource "aws_ecr_repository" "heartbeat" {
+  # The :latest tag is used in Staging
+  #tfsec:ignore:AWS078
```
patheard, line 3: You should be able to remove this comment since the warning will get suppressed below by this one:
```hcl
image_tag_mutability = "MUTABLE" #tfsec:ignore:AWS078
```

aws/heartbeat/lambda.tf
```diff
@@ -0,0 +1,88 @@
...
+  timeout = 60
+
+  memory_size = 1024
+  publish     = true
+
+  lifecycle {
+    ignore_changes = [
+      image_uri,
+    ]
+  }
```
patheard, line 37: If you're always using the `latest` tag for the image, you can remove this lifecycle rule.  Normally you'd only have an ignore on the `image_uri` if it was being updated by another process outside of Terraform.

aws/heartbeat/lambda.tf
```diff
@@ -0,0 +1,88 @@
...
+        Sid    = ""
+        Principal = {
+          Service = "lambda.amazonaws.com"
+        }
+      },
+    ]
+  })
+}
+
+resource "aws_lambda_function" "heartbeat" {
```
patheard, line 21: We have a Terraform Lambda module that'll take care of a lot of the setup for you here:
https://github.com/cds-snc/terraform-modules/tree/main/lambda

If you want to see an example of it in action, we've got one in the [ScanFiles `lambda.tf`](https://github.com/cds-snc/scan-files/blob/main/terragrunt/aws/api/lambda.tf#L1-L32).

Also happy to help you get it up and running if you get stuck!

aws/heartbeat/lambda.tf
```diff
@@ -0,0 +1,88 @@
...
+  source_arn    = aws_lambda_function.heartbeat.arn
+  qualifier     = aws_lambda_alias.heartbeat-lambda-alias.name
+}
+
+resource "aws_lambda_alias" "heartbeat-lambda-alias" {
+  name             = "heartbeat-lambda-alias"
+  description      = "Heartbeat testing lambda alias"
+  function_name    = aws_lambda_function.heartbeat.function_name
+  function_version = "$LATEST"
+}
```
patheard, line 87: I think you can get away without using an `aws_lambda_alias` here.  Usually this is for when you want to maintain multiple versions of a Lambda function and switch easily between them.

env/staging/heartbeat/terragrunt.hcl
```diff
@@ -0,0 +1,13 @@
...
+
+inputs = {
+  billing_tag_key                             = "CostCenter"
+  billing_tag_value                           = "notification-canada-ca-staging"
+  schedule_expression                         = "cron(*/1 * * * *)"
+}
+
+terraform {
+  source = "../../../aws/heartbeat"
```
patheard, line 12: This needs a double slash to help resolve the path:
```hcl
source = "../../../aws//heartbeat"
```

aws/heartbeat/lambda.tf
```diff
@@ -0,0 +1,88 @@
...
+    variables = {
+      TF_VAR_heartbeat_api_key                    = var.heartbeat_api_key
+      TF_VAR_heartbeat_base_url                   = var.heartbeat_base_url
+      TF_VAR_heartbeat_template_id                = var.heartbeat_template_id
+    }
+  }
+}
+
+resource "aws_cloudwatch_event_target" "heartbeat" {
+  arn  = aws_lambda_function.heartbeat-test.arn
```
patheard, line 49: I think this should just be `aws_lambda_function.heartbeat.arn` to reference the Lambda function.

patheard (changes requested):
Alrighty, added some fixes/comments, but overall it's a good start!

env/staging/heartbeat/terragrunt.hcl
jzbahrai: I originally put it as that but was confused about the double slash.

aws/heartbeat/lambda.tf
jzbahrai: ya I think I am going to need help with that hahaha. Will set up something.

env/staging/heartbeat/terragrunt.hcl
patheard: Totally agree - it is confusing.  The reason behind it is related to how Terraform references modules in sub-directories:
https://www.terraform.io/docs/language/modules/sources.html#modules-in-package-sub-directories

Terragrunt is just using the same rules as Terraform and uses the `//` to indicate that everything before it (`../../../aws`) is the module source and the big after is a sub-directory within that module.

env/staging/heartbeat/terragrunt.hcl
jzbahrai: done

aws/heartbeat/ecr.tf
jzbahrai: done

aws/heartbeat/lambda.tf
jzbahrai: done

aws/heartbeat/lambda.tf
```diff
@@ -0,0 +1,88 @@
...
+    input_template = <<-EOF
+    {
+      "instance_id": <instance>,
+      "instance_status": <status>
+    }
+    EOF
+  }
+}
+
+resource "aws_cloudwatch_event_rule" "heartbeat-testing" {
```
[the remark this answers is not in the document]
jzbahrai: done

env/staging/heartbeat/terragrunt.hcl
jzbahrai: done.

aws/heartbeat/lambda.tf
jzbahrai: changed.

env/production/heartbeat/terragrunt.hcl
```diff
@@ -0,0 +1,13 @@
+include {
+  path = find_in_parent_folders()
+}
+
+inputs = {
+  billing_tag_key                             = "CostCenter"
+  billing_tag_value                           = "notification-canada-ca-production"
```
patheard, line 7: You'll need to define this variable in your `aws/heartbeat/variables.tf`.  That's why the `terraform plan` is complaining.

env/production/heartbeat/terragrunt.hcl
```diff
@@ -0,0 +1,13 @@
+include {
+  path = find_in_parent_folders()
+}
+
+inputs = {
+  billing_tag_key                             = "CostCenter"
```
patheard, line 6: I think you can get rid of this - I don't see it being used anywhere.

env/staging/heartbeat/terragrunt.hcl
```diff
@@ -0,0 +1,13 @@
+include {
+  path = find_in_parent_folders()
+}
+
+inputs = {
+  billing_tag_key                             = "CostCenter"
+  billing_tag_value                           = "notification-canada-ca-staging"
```
patheard, line 7: Same comments as the prod `terragrunt.hcl`

patheard (changes requested):
Just a few gotchas with variables that need to be defined.

You'll also need to update the merge workflows with a block to `terraform apply` the heartbeat lambda changes.  Something like so for each one:

* [Prod](https://github.com/cds-snc/notification-terraform/blob/main/.github/workflows/merge_to_main_production.yml#L117-L120)
* [Staging](https://github.com/cds-snc/notification-terraform/blob/main/.github/workflows/merge_to_main_staging.yml#L125-L128)

jzbahrai:
@patheard made all the changes :) thank you for the thorough review.

sastels:
> @sastels yup done in this PR: #338
> also made the updates we talked about.

Oh, awesome! 🙃

sastels:
In the terraform plan I see
```
  + "heartbeat_template_id" = "\"\""
```
Is the template I’d not being set correctly somewhere?

patheard:
> In the terraform plan I see
> 
> ```
>   + "heartbeat_template_id" = "\"\""
> ```
> 
> Is the template I’d not being set correctly somewhere?

Good catch @sastels.  Looks like the value is coming from a [GitHub secret](https://github.com/cds-snc/notification-terraform/blob/main/.github/workflows/merge_to_main_staging.yml#L24).

patheard (approved):
I'd say it's good to merge once that `heartbeat_template_id` is figured out!

sastels (approved):
Let's try this out!

PR #341, opened 2021-12-07
WAF rules for valid endpoints of notification-document-download-api

jimleroyer:
# Summary | Résumé

Added WAF rules that filter in only valid endpoints for document download API.

* I will do some clicksops to make sure the regex actually works.

      langues officielles?
      fonctionnalité existante?


aws/eks/waf.tf
```diff
@@ -118,6 +118,46 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
     }
   }
```
jimleroyer, line 120: WAF rules in the eks module seems a bit off. I wonder if that wouldn't be best to move them in a better location. Any suggestion?

aws/eks/waf.tf
```diff
@@ -128,3 +168,32 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+
+  # GET /services/<uuid:service_id>/documents/<uuid:document_id>
+  regular_expression {
+    regex_string = "/services/[\\w]{8}-[\\w]{4}-[\\w]{4}-[\\w]{4}-[\\w]{12}/documents/[\\w]{8}-[\\w]{4}-[\\w]{4}-[\\w]{4}-[\\w]{12}"
+  }
+
+  # POST /services/<uuid:service_id>/documents
+  regular_expression {
+    regex_string = "/services/[\\w]{8}-[\\w]{4}-[\\w]{4}-[\\w]{4}-[\\w]{12}/documents"
+  }
```
jimleroyer, line 194: I'll need more regex to allow api and admin components.

aws/eks/waf.tf
```diff
@@ -128,3 +168,32 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+  # https://docs.aws.amazon.com/waf/latest/developerguide/waf-regex-pattern-set-managing.html
+
+  # GET /_status
+  regular_expression {
+    regex_string = "/_status"
+  }
+
+  # GET /services/<uuid:service_id>/documents/<uuid:document_id>
+  regular_expression {
+    regex_string = "/services/[\\w]{8}-[\\w]{4}-[\\w]{4}-[\\w]{4}-[\\w]{12}/documents/[\\w]{8}-[\\w]{4}-[\\w]{4}-[\\w]{4}-[\\w]{12}"
```
patheard, line 188: I'm not 100% sure if you need to escape the `\w` first slash.  It might just be:
 ```hcl
regex_string = "/services/[\w]{8}-[\w]{4}-[\w]{4}-[\w]{4}-[\w]{12}/documents/[\w]{8}-[\w]{4}-[\w]{4}-[\w]{4}-[\w]{12}"
```

**Review by patheard — APPROVED** [review]
Looks good!  I left a comment about escaping for the `\w` leading slash, but I'm not sure which is correct so I'd say go for a merge and we can see what happens in staging.

File: aws/eks/waf.tf (hunk 3 above)
Comment (jimleroyer) [reply aws/eks/waf.tf:188]: I tried and it kept complaining that this was an invalid escape character. Browsing some similar TF issues, I saw double backslashes might be necessary. It started to work then. 🤷

File: aws/eks/waf.tf [hunk 4]
```diff
@@ -118,6 +118,130 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+        statement {
+          or_statement {
+
+            statement {
+              and_statement {
+                statement {
+                  byte_match_statement {
+                    positional_constraint = "STARTS_WITH"
+                    field_to_match {
+                      uri_path {}
```
jimleroyer, line 144: I don't think this statement will work unfortunately. There is no field to match on the hostname, I am puzzled to the reason why not. That would make the rule a lot easier to write. It seems I will have to create rules for all of admins and api if my only option is to filter through the paths.

patheard: You could try using an [HTTP Host header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host) match statement:
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl#single-header

That would allow you to filter on the document download API hostname requests only (in theory).  I haven't tried this type of statement before but it looks promising.

aws/eks/waf.tf
```diff
@@ -118,6 +118,130 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+              }
+            }
+
+            statement {
+              byte_match_statement {
+                positional_constraint = "STARTS_WITH"
+                field_to_match {
+                  uri_path {}
+                }
+                search_string = "http://api"
```
patheard, line 183: As you suspected, this isn't going to work unfortunately since the hostname isn't included in the `uri_path`.

aws/eks/waf.tf
```diff
@@ -118,6 +118,130 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+
+            statement {
+              or_statement {
+                statement {
+                  byte_match_statement {
+                    positional_constraint = "CONTAINS"
+                    field_to_match {
+                      uri_path {}
+                    }
+                    search_string = "staging.notification.cdssandbox.xyz"
```
patheard, line 203: Same comment re: hostname in the `uri_path`.

aws/eks/waf.tf
jimleroyer: Ohhh true! 💥 Thank you!! Will try to include this change this pm!

aws/eks/waf.tf
```diff
@@ -118,6 +118,138 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+
+    action {
+      block {
+        custom_response {
+          response_code = 204
+        }
+      }
+    }
+
+    statement {
```
patheard, line 133: I suspect this will fail to `terraform apply` because you have 4 levels of nested `statement` blocks:
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl#statement
> Although the statement block is recursive, currently only 3 levels are supported

aws/eks/waf.tf
```diff
@@ -118,6 +118,138 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+                }
+                text_transformation {
+                  priority = 2
+                  type     = "LOWERCASE"
+                }
+              }
+            }
+
+            statement {
+              or_statement {
```
patheard, line 200: I'd recommend setting the `search_string` here as a variable so you didn't need the `or_statement` to check for both staging/prod.

aws/eks/waf.tf
jimleroyer: Arff...

jimleroyer: The not `!` operator itself requires a `statement`, hence this block for #2 would lead to 4 `statement` blocks overall unfortunately. There is no negation in support [in `byte_match_statement` unfortunately](https://docs.aws.amazon.com/waf/latest/APIReference/API_ByteMatchStatement.html), i.e. `DONT_STARTS_WITH` or something similar.

```
statement {
  or_statement { # 1 or 2?
    # at 2... first option let's say with !STARTS_WITH "API"
    statement {
      or_statement {
        statement {
          not_statement {
            statement {
               # STARTS_WITH "api"...
               byte_match_statement { ... }
            }
          }
        }
      }
    }
  }
}
```

I will try with a regex though and see if we can cheat.

I wonder, is the 3 level nested block limit a constraint by AWS or Terraform? I didn't see that in AWS documentation so far.

aws/eks/waf.tf
```diff
@@ -118,6 +118,89 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+                text_transformation {
+                  priority = 1
+                  type     = "COMPRESS_WHITE_SPACE"
+                }
+                text_transformation {
+                  priority = 2
+                  type     = "LOWERCASE"
+                }
+              }
+            }
```
patheard, line 172: I think this needs to be wrapped in a `not_statement` since you only want to block if the `aws_wafv2_regex_pattern_set.re_document_download` does not match the `uri_path`. WAF ACL Terraform syntax makes my 🧠  hurt...

aws/eks/waf.tf
patheard: Pretty sure it's a Terraform constraint, but I believe it only applies to `statement` blocks, so things like `or_statement`, `and_statement`, `not_statement` don't count towards the 3 levels.

jimleroyer: Yes it's my understanding as well, i.e. only `statement` is counted. I wish TF did not have this artificial constraint... 😞

aws/eks/waf.tf
jimleroyer: Did you notice the negative assertion in the regex though? That would act as the `!` operator.

jimleroyer: Spoke to Pat on another medium. He really meant that the negation should be applied on the statement at line 157 but I assumed he meant on the one at line 176.

aws/eks/waf.tf
jimleroyer: @patheard just split in two the rule as agreed and we get rid of a `statement` level. It also has the added benefit of defining priority so we can enforce which expression to evaluate first (document download api and then more generic api calls in this instance).

aws/eks/waf.tf
```diff
@@ -118,6 +118,108 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+
+    action {
+      block {
+        custom_response {
+          response_code = 204
+        }
+      }
+    }
+
+    statement {
```
patheard, line 197: I like this as the 2 rules now since it means if the regex `!` negation doesn't work, you can just wrap this in a `not_statement`.

patheard (approved):

aws/eks/waf.tf
jimleroyer: Indeed, it provides us a bit more space for more complex rules. 👍

aws/eks/waf.tf
sastels: this WAF is just in front of the eks, so it's not terrible to have it here. There's another WAF for cloudfront over in that module.

sastels (approved):
Nice, I like how that `STARTS_WITH` block allows us to scope down to the download api server and not have to do all of them at once 👍 

Let's see if this works...

PR #333, opened 2021-12-01
Added alarms for lambda to cloudwatch

jzbahrai:
# Summary | Résumé

Add alerts for when 500 errors appear in the API lambda logs

# Test instructions | Instructions pour tester la modification

ran terragrunt tf-plan

PR #348, opened 2021-12-13
add variables for production

jzbahrai:
We were missing production variables from here: https://github.com/cds-snc/notification-terraform/pull/333/files
Adding them in this PR.


sastels (approved):
🚀

PR #352, opened 2021-12-14
bug/Checkov issues with api gateway

sastels:
# Summary | Résumé

https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/cds-snc/notification-planning/279

- add logging level
- put a waf in front, currently with the same rules as the eks' waf

Decided to use a new waf rather than just use the eks' waf - this will make things simpler if we want to add specific rules for different servers (ie whitelisting list of endpoints)

      langues officielles?
      fonctionnalité existante?

### DIFF
--- a/aws/lambda-api/api_gateway.tf
+++ b/aws/lambda-api/api_gateway.tf
@@ -123,3 +123,14 @@ resource "aws_api_gateway_base_path_mapping" "api" {
   stage_name  = aws_api_gateway_stage.api.stage_name
   domain_name = aws_api_gateway_domain_name.api.domain_name
 }
+
+resource "aws_api_gateway_method_settings" "all" {
+  rest_api_id = aws_api_gateway_rest_api.api.id
+  stage_name  = aws_api_gateway_stage.api.stage_name
+  method_path = "*/*"
+
+  settings {
+    metrics_enabled = true
+    logging_level   = "INFO"
+  }
+}
--- a/aws/lambda-api/api_gateway.tf
+++ b/aws/lambda-api/api_gateway.tf
@@ -131,7 +131,7 @@ resource "aws_api_gateway_method_settings" "all" {
 
   settings {
     metrics_enabled = true
-    logging_level   = "INFO"
+    logging_level   = "ERROR"
   }
 }
 
--- a/aws/lambda-api/api_gateway.tf
+++ b/aws/lambda-api/api_gateway.tf
@@ -137,5 +137,5 @@ resource "aws_api_gateway_method_settings" "all" {
 
 resource "aws_wafv2_web_acl_association" "waf_association" {
   resource_arn = aws_api_gateway_stage.api.arn
-  web_acl_arn  = var.waf_arn
+  web_acl_arn  = aws_wafv2_web_acl.api_lambda.arn
 }
--- a/aws/lambda-api/api_gateway.tf
+++ b/aws/lambda-api/api_gateway.tf
@@ -134,3 +134,8 @@ resource "aws_api_gateway_method_settings" "all" {
     logging_level   = "INFO"
   }
 }
+
+resource "aws_wafv2_web_acl_association" "waf_association" {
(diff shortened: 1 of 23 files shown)


aws/lambda-api/api_gateway.tf
```diff
@@ -123,3 +123,19 @@ resource "aws_api_gateway_base_path_mapping" "api" {
...
 }
+
+resource "aws_api_gateway_method_settings" "all" {
+  rest_api_id = aws_api_gateway_rest_api.api.id
+  stage_name  = aws_api_gateway_stage.api.stage_name
+  method_path = "*/*"
+
+  settings {
+    metrics_enabled = true
+    logging_level   = "INFO"
```
patheard, line 134: Do you want to leave `INFO` logging level in staging/production?  It'll likely be chatty.

aws/lambda-api/waf.tf
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  # Use a bunch of AWS managed rules
+  # See https://docs.aws.amazon.com/waf/latest/developerguide/aws-managed-rule-groups-list.html
+  rule {
+    name     = "AWSManagedRulesAmazonIpReputationList"
+    priority = 1
+
+    override_action {
+      count {}
```
patheard, line 16: Do you want these all set to `count`?  This will only count requests that match, but still allow the traffic through. 
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl#override-action

patheard (commented):
Just had some questions around logging level and count vs. block in the WAF ACL.

One other rule that might be handy is an [IP based rate limit](https://github.com/cds-snc/gc-articles/blob/main/infrastructure/terragrunt/aws/load-balancer/waf.tf#L361-L381) as well (if it makes sense).

aws/lambda-api/waf.tf
jimleroyer: Seems like this is what we have in our EKS config at the moment:
https://github.com/cds-snc/notification-terraform/blob/bcbf0a715e4690e6da77ef1c9de4c47c728c462c/aws/eks/waf.tf#L15

jimleroyer:
Heads up to tell we'll have to include the new rules defined in there once tested in staging:
https://github.com/cds-snc/notification-terraform/pull/341

aws/lambda-api/waf.tf
patheard: Yes, since it means that right now, none of the WAF rules are blocking any requests.   That being said, it would be good to get to the root of why `count` was set in EKS.   It's possibly because some rules were breaking Notify behaviour, but there are ways we can selectively filter out the rules we don't want to apply.

sastels: It's unclear - we used `count` since the start of using a waf
https://github.com/cds-snc/notification-terraform/pull/44

aws/lambda-api/api_gateway.tf
sastels: 👍  I can switch to `ERROR` if that will be more useful (levels are `OFF`, `INFO`, and `ERROR`)

sastels:
> One other rule that might be handy is an IP based rate limit as well (if it makes sense).

I suspect a rate limit might block heavy users like HC? thoughts, @jimleroyer ?

aws/lambda-api/waf.tf
patheard: Yeah, I was looking at that.  My suspicion is that it was originally setup this way to see how the rules would behave but they should be switched over to `block` by changing to:
```hcl
override_action {
    none {}
}
```
Before doing this, we'd definitely have to see which rules are being triggered for valid requests and skip those cases.

sastels:
@patheard @jimleroyer I believe I've added the recent changes to the EKS WAF here as well.

patheard (approved):

jimleroyer (approved):
:shipit:

PR #353, opened 2021-12-14
Checkov api lambda iam issues

sastels:
# Summary | Résumé

https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/cds-snc/notification-planning/278

- moved s3 action to its own statement and restricted to the csv upload bucket
- deleted PinPoint and SES actions
- had issues with the plan not picking out the new `common` outputs, so upgraded terragrunt and used the `mock_outputs_merge_with_state` attribute (as discussed below). Unclear if we need to do this in prod.

Does api need `sns:Publish` and `securityhub:BatchImportFindings` ?

      langues officielles?
      fonctionnalité existante?


.github/workflows/terragrunt_plan_staging.yml
```diff
@@ -14,7 +14,7 @@ env:
   AWS_SECRET_ACCESS_KEY: ${{ secrets.STAGING_AWS_SECRET_ACCESS_KEY }}
   AWS_REGION: ca-central-1
   TERRAFORM_VERSION: 0.14.4
-  TERRAGRUNT_VERSION: 0.31.0
+  TERRAGRUNT_VERSION: 0.35.13
```
sastels, line 17: needed for `mock_outputs_merge_with_state` attribute

.github/workflows/terragrunt_plan_staging.yml
```diff
@@ -72,7 +72,7 @@ jobs:
           filters: |
             common:
               - '.github/workflows/terragrunt-plan-staging.yml'
-              - 'env/common/**'
+              - 'aws/common/**'
```
sastels, line 75: typo

aws/lambda-api/iam.tf
```diff
@@ -47,14 +47,18 @@ data "aws_iam_policy_document" "api_policies" {
   statement {
     effect = "Allow"
     actions = [
-      "mobiletargeting:*",
-      "ses:SendEmail",
-      "ses:SendRawEmail",
```
sastels, line 52: Deleted these 3 actions (in k8s api and celery have the same iam policy, this was just copied from that). Note that the `mobiletargeting` one is for PinPoint.

.github/workflows/terragrunt_plan_staging.yml
patheard: good catch on the missing `aws/common/**`, but I think you'll also want `env/common/**` to catch changes to the `terragrunt.hcl` file in there.

patheard (commented):
Looks good Steve!  I just had a question about keeping `env/common/**` in the path filter checking for changes.

I was also curious about this in the plan:

```hcl
  # module.notify_slack_critical.module.lambda.null_resource.archive[0] must be replaced
-/+ resource "null_resource" "archive" {
      ~ id       = "8695568181817461827" -> (known after apply)
      ~ triggers = { # forces replacement
          ~ "timestamp" = "1639579846555974000" -> "<WARNING: Missing lambda zip artifacts wouldn't be restored>"
            # (1 unchanged element hidden)
        }
    }
```

It seems to be related to the `terraform-aws-modules/notify-slack/aws` module using an old version of the `terraform-aws-modules/lambda/aws` module:

terraform-aws-modules/terraform-aws-lambda#179

.github/workflows/terragrunt_plan_staging.yml
sastels: ah, yeah, there's no `env/common` actually, but probably the typo was actually supposed to be `env/staging/common/**` 👍

patheard (approved):

sastels:
> It seems to be related to the terraform-aws-modules/notify-slack/aws module using an old version of the terraform-aws-modules/lambda/aws module:

I don't know if this is true - we're using the latest version of `notify-slack` which uses the version of `lambda` from Dec 9
https://github.com/cds-snc/notification-terraform/runs/4538085828?check_suite_focus=true#step:6:50

patheard:
> > It seems to be related to the terraform-aws-modules/notify-slack/aws module using an old version of the terraform-aws-modules/lambda/aws module:
> 
> I don't know if this is true - we're using the latest version of `notify-slack` which uses the version of `lambda` from Dec 9 https://github.com/cds-snc/notification-terraform/runs/4538085828?check_suite_focus=true#step:6:50

Ah, you're totally right.  When I looked at the module source, I had misread this line in the source as an older version:
https://github.com/terraform-aws-modules/terraform-aws-notify-slack/blob/master/main.tf#L73

PR #354, opened 2021-12-14
Update lambda config for failure

jzbahrai:
# Summary | Résumé
Update lambda config: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function_event_invoke_config .

Goal is to reduce invocation on failure.

### DIFF
--- a/aws/heartbeat/lambda.tf
+++ b/aws/heartbeat/lambda.tf
@@ -15,6 +15,12 @@ module "heartbeat" {
   }
 }
 
+resource "aws_lambda_function_event_invoke_config" "heartbeat_" {
+  function_name                = module.heartbeat.function_name
+  maximum_event_age_in_seconds = 60
+  maximum_retry_attempts       = 0
+}
+
 resource "aws_cloudwatch_event_target" "heartbeat" {
   arn  = module.heartbeat.function_arn
   rule = aws_cloudwatch_event_rule.heartbeat_testing.id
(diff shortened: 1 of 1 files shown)


aws/heartbeat/lambda.tf
```diff
@@ -15,6 +15,12 @@ module "heartbeat" {
   }
 }
 
+resource "aws_lambda_function_event_invoke_config" "heartbeat_" {
+  function_name                = module.heartbeat.function_name
```
jimleroyer, line 19: Just tagging @patheard to know if that will work via the module this way. The plan seems to work and define some changes.

jimleroyer (commented):
LGTM, just want Pat to confirm!

aws/heartbeat/lambda.tf
```diff
@@ -15,6 +15,12 @@ module "heartbeat" {
   }
 }
 
+resource "aws_lambda_function_event_invoke_config" "heartbeat_" {
```
patheard, line 18: nit: just around naming, I'd either remove the trailing `_` from the `heartbeat_` resource name or call it something like `heartbeat_invoke_config`.  Totally a preference thing though.

patheard (approved):
Look good!  Left a minor non-blocking comment but I'm curious to see if this works as advertised.

aws/heartbeat/lambda.tf
jzbahrai: this was an error on my part 🤦🏽 I meant to write a whole word.

PR #356, opened 2021-12-15
Checkov lambda issues

sastels:
# Summary | Résumé

https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/cds-snc/notification-planning/281

Fix a few issues that Checkov identified:
- set a reserved concurrency (to 850). This means that there will always be 850 lambda slots available for us to use for the api (there might not be that many functions running if the system is quiet, but there will be there if needed). It also means that there is a cap on how many lambda resources the api will use (out of the 1000 for the account) so api cannot grow so large as to prevent other lambdas (ex: slack notifications) from running.
- turn on x-ray tracing. This is a way to investigate lambda failures and performance issues. Sounds like something that might be useful at least until we're confident that everything is configured correctly.

      langues officielles?
      fonctionnalité existante?


patheard (approved):

aws/lambda-api/lambda.tf
```diff
@@ -1,14 +1,18 @@
...
-  role    = aws_iam_role.api.arn
-  timeout = 60
+  timeout                        = 60
+  reserved_concurrent_executions = 850
+  memory_size                    = 1024
 
-  memory_size = 1024
-  publish     = true
+  tracing_config {
+    mode = "Active"
```
jimleroyer, line 14: What kind of additional logs could that provide to us? Could we resolve a bit more the new relic integration issues with it? (I assume not but I wish -- I can't see logs for new relic failing; we might have to turn on a bit more logging).

jimleroyer (approved):
Thank you for the explanation in the PR and providing key information on lambda constraints, it's quite helpful. 👍

aws/lambda-api/lambda.tf
sastels: I'm not sure - I figure we turn this on and see if it's useful and turn it off if not 🤷‍♂️ 
Might be useful if it gives us some performance insights.

PR #358, opened 2021-12-15
fix: WAF block based on inputs and IP reputation

patheard:
# Summary
Enable request blocking based on requests that have known bad input
(e.g. the Log4j exploit) or are from IP addresses with a poor
reputation score.

A subsequent PR will be created to enable logging so we can determine
which additional AWS managed rules can be set to block as well.

# Related

### DIFF
--- a/aws/eks/waf.tf
+++ b/aws/eks/waf.tf
@@ -13,7 +13,7 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
     priority = 1
 
     override_action {
-      count {}
+      none {}
     }
 
     statement {
@@ -57,7 +57,7 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
     priority = 3
 
     override_action {
-      count {}
+      none {}
     }
 
     statement {


jzbahrai (commented):
its that simple? lgtm but I dont know what I am approving so will wait for someone else.

patheard:
> its that simple? lgtm but I dont know what I am approving so will wait for someone else.

Yup, it's definitely an "easy change, high impact" kind of situation 😅   

If we set all the managed rules to block we'd definitely mess with some legitimate Notify requests because of things like large request body size or missing user agent headers.

aws/eks/waf.tf
```diff
@@ -13,7 +13,7 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
     priority = 1
 
     override_action {
-      count {}
+      none {}
```
jimleroyer, line 16: A none overriding action means we would rely on the default allow action, no? I would expect [a `block` block](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl#block) if we want to stop the bots identified by this rule.

aws/eks/waf.tf
```diff
@@ -57,7 +57,7 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
     priority = 3
 
     override_action {
-      count {}
+      none {}
```
jimleroyer, line 60: Same here, but maybe I misread the TF/AWS documentation?

aws/eks/waf.tf
patheard: It's a bit confusing - when you use a managed ruleset, you only specify [an `override_action`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl#override-action), which can only have a value of `count` or `none`.  In the `none` case, it means use the managed ruleset's default (`block` in these cases).

jimleroyer: Definitely confusing, I guess there is no way to know the ruleset's default [other than AWS documentation](https://docs.aws.amazon.com/waf/latest/developerguide/aws-managed-rule-groups-list.html)?

jimleroyer (approved):

PR #361, opened 2021-12-15
Update cloudwatch alarms for heartbeat

jzbahrai:
# Summary | Résumé

Add cloudwatch alarms for the new heartbeat api

### DIFF
--- a/aws/heartbeat/cloudwatch_logs.tf
+++ b/aws/heartbeat/cloudwatch_logs.tf
@@ -1,0 +1,53 @@
+#
+# Heartbeat CloudWatch logging
+#
+
+resource "aws_cloudwatch_log_group" "heartbeat_log_group" {
+  name              = "heartbeat_log_group"
+  retention_in_days = 90
+  tags = {
+    CostCenter  = "notification-canada-ca-${var.env}"
+    Environment = var.env
+    Application = "lambda"
+  }
+}
+
+# This account will be used by Heartbeat resources in the account and region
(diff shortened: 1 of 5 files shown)


aws/heartbeat/cloudwatch_logs.tf
```diff
@@ -0,0 +1,53 @@
...
+    principals {
+      type        = "Service"
+      identifiers = ["apigateway.amazonaws.com"]
+    }
+  }
+}
+
+resource "aws_cloudwatch_log_metric_filter" "heartbeat-500-errors-api" {
+  name           = "heartbeat-500-errors-api"
+  pattern        = "\"\\\" 500 \""
```
jzbahrai, line 45: @sastels based on your PR does this need to be [500]

sastels: Possibly just `[ERROR]` actually. ie to catch if the heartbeat crashes for some reason.

jimleroyer: I raised the question with Jumana. @jzbahrai Did you look in the  logs with 500s what will be the error format? To make sure the failed heartbeat will really go through the CW metric filter?

I know that @sastels had to change the one in API lambda hence I was wondering.

jimleroyer (commented):
This is great overall! good job 💪 

I just wonder if the logs output for the heartbeat lambda will really go through the CW metric filter. I'll wait for Steve's take on it.

sastels (approved):
I think this will work! 😅

PR #363, opened 2021-12-16
feat: send EKS WAF logs to a CloudWatch log group

patheard:
# Summary
Add CloudWatch log group to capture the EKS WAF ACL logs.  These logs
will be used to determine which managed rulesets can be set to `block`.

# Related


patheard:
Note that the plan has resources from #341 that failed to create.  Will wait for a fix to that before merging.

jimleroyer (approved):
LGTM

PR #366, opened 2021-12-16
feat: add WAF Bot control managed rules

patheard:
# Summary
Add AWS's Bot control managed rules to the EKS WAF ACL.  These will
be initially set to `count` to determine impact, and then switched over
to `block`.

# Related


jimleroyer (approved):
LGTM

PR #369, opened 2022-01-10
Update cloudwatch_alarms.tf

jzbahrai:
# Summary | Résumé

We are changing the evaluation to be period for 1 min over a 10 minute period.


sastels (approved):
👍

PR #373, opened 2022-01-17
Added queue alarms around database tasks

jimleroyer:
# Summary | Résumé

Added an alert around the db tasks operations, including saving ones.

Let's push to staging environment and see if the alert works correctly.


sastels (approved):
nice!

PR #374, opened 2022-01-18
Removed alert around admin looking at sensitive services

jimleroyer:
# Summary | Résumé

Due to the sheer amount of alerts we get and how these alerts are not actionables for now, we decided to remove the alerts around admins looking at sensitive services. The important part for us is to log it so that we can trace it back if necessary.

[A new Zenhub card was created](https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/cds-snc/notification-planning/325) to surface this information in a more effective way.

We'll know soon enough if that works or not. 😅


sastels (approved):
yeah, this is probably for the best. :shipit:

PR #376, opened 2022-01-19
Reduce noise on the database-tasks critical alert 

jimleroyer:
# Summary | Résumé

We currently have the database-tasks queue being the bottleneck and its related critical alert easily gets triggered as we are not operating under Notify's SLOs. We are working to improve this but in the meantime, we will modify the alert to level with current Notify's performance expectations.


sastels (approved):
Let's see if this cuts the noise

sastels:
Should we bump up the warning as well? It's triggering a lot
image

jimleroyer:
I also increased severity threshold from 1 minute to 5 minutes for the warning on db-tasks queue after speaking to Steve.

PR #380, opened 2022-02-11
Block requests identified in the linux rule group

jimleroyer:
# Summary | Résumé

We got a surge of PHP and github scans in the last week, increasing our suppor time. After a review of last week requests made and how it AWS WAF ACL rules would block or not these requests, it seems quite a fit for us to catch a break.

### DIFF
--- a/aws/eks/waf.tf
+++ b/aws/eks/waf.tf
@@ -78,10 +78,6 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
     name     = "AWSManagedRulesLinuxRuleSet"
     priority = 4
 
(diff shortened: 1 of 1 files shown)


aws/eks/waf.tf
```diff
@@ -35,7 +35,7 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
     priority = 2
 
     override_action {
-      count {}
+      block {}
```
jimleroyer, line 38: Oopsie, wrong rule set, let me change that.

patheard (approved):
🛑

PR #385, opened 2022-02-16
feat: post Slack message if Terraform apply fails

patheard:
# Summary
Message will be posted to the Notify DEV channel:
image

This is a stop-gap until we've had time to update the Terraform provider to fix the issues with creating and updating resources like the WAF ACL and Lambda API function.

# Related
* cds-snc/notification-planning#411


jimleroyer (approved):
Thanks Pat for making our process better.

mohamed-cds (approved):
Simple, yet effective! 🚀 ✅

PR #386, opened 2022-02-17
chore: bump Terraform and Terragrunt versions

patheard:
# Summary
Upgrade to Terraform 1.1.5 and Terragrunt 0.36.1.  This is to resolve errors we're seeing with the older version of Terraform not waiting for resources to be ready before attempting to configure them.

## Also included:

* Pins Terraform dependencies to ensure more consistent Terraform plan/apply runs.
* Removes the duplicated AWS API gateway account resource.  There can only be one of these set per AWS account
and the Lambda API creation of this resource will work as expected.

# Related
* cds-snc/notification-planning#411


.github/workflows/terragrunt_plan_staging.yml
```diff
@@ -13,8 +13,8 @@ env:
   AWS_ACCESS_KEY_ID: ${{ secrets.STAGING_AWS_ACCESS_KEY_ID }}
   AWS_SECRET_ACCESS_KEY: ${{ secrets.STAGING_AWS_SECRET_ACCESS_KEY }}
   AWS_REGION: ca-central-1
-  TERRAFORM_VERSION: 0.14.4
-  TERRAGRUNT_VERSION: 0.35.13
+  TERRAFORM_VERSION: 1.1.5
+  TERRAGRUNT_VERSION: 0.36.1
```
jimleroyer, line 17: Hmm many different places to configure this. No way to define it in only one place I assume?

patheard: We could try something like a `.env` file and then parse it into the GitHub workflows like so:
https://arinco.com.au/blog/github-actions-share-environment-variables-across-workflows/

May even be possible to use the same `.env` for the devcontainer.json so we'd only need it defined once in the project.

patheard: Added a `.env` with versions that are used for all workflows.  Unfortunately I didn't find a way to use this for the `devcontainer.json`.

jimleroyer:
env is very nice touch, indeed! 🥳

PR #389, opened 2022-02-22
feat: lower Lambda API provisioned concurrency

patheard:
# Summary 
Provisioned concurrency has an associated cost. In Production, 
since the Lambda API is currently only handling 5% of requests, 
the min capacity can be scaled down to 1.  

Once we increase the traffic to the Lambda API, this value can 
be scaled up.

# ⚠️  Note
There should be no Staging changes for this PR.


mohamed-cds (approved):
makes sense. 🚀 ✅

PR #396, opened 2022-02-25
feat: reduce noise in pr terraform plan outputs

mohamed-cds:
Reduce the amount of unnecessary outputs when reviewing terraform changes, specifically the ecr iam policy. This will also determine whether the lambda hashes are correctly generated using the updated aws slack notification module.

- Use locals to reference lambda api arns ([explanation on why this helps](https://github.com/hashicorp/terraform/issues/27171#issuecomment-740249394))
- Upgrade module to ~> 4.24.0 so that `TF_RECREATE_MISSING_LAMBDA_PACKAGE` is used by the underlying module


sastels (approved):
LGTM. 
Interesting about the local usage.
You meant upgrade module to 4.24.0 I assume.

PR #404, opened 2022-03-03
chore: reintroduce lambda package recreate flag

mohamed-cds:
New version of the lambda and slack modules still create unnecessary changes. Bring back the flag that will disable the recreate of the slack lambda.


.github/workflows/terragrunt_plan_staging.yml
```diff
@@ -41,6 +41,9 @@ env:
   TF_VAR_api_domain_name: ${{ secrets.STAGING_API_DOMAIN_NAME }}
   TF_VAR_api_lambda_domain_name: ${{ secrets.STAGING_API_LAMBDA_DOMAIN_NAME }}
   TF_VAR_new_relic_license_key: ${{ secrets.STAGING_NEW_RELIC_LICENSE_KEY }}
+  # Prevents repeated creation of the Slack lambdas if already existing.
+  # See: https://github.com/terraform-aws-modules/terraform-aws-notify-slack/issues/84
+  TF_RECREATE_MISSING_LAMBDA_PACKAGE: false
```
patheard, line 46: Can you try adding this to `.github/workflows/merge_to_main_staging.yml` as well?  I'm curious if it needs to be added to the `terraform apply` to actually get rid of the plan flip/flop.

patheard (approved):

PR #408, opened 2022-03-09
feat: differentiate ok actions and update alarm thresholds

mohamed-cds:
This PR makes it easier to distinguish ok actions and alarms. Production thresholds for the newly introduced inflight and bulk buffer metrics are being increased to exceed `normal` traffic

### DIFF
--- a/aws/common/cloudwatch_alarms.tf
+++ b/aws/common/cloudwatch_alarms.tf
@@ -46,7 +46,7 @@ resource "aws_cloudwatch_metric_alarm" "sns-spending-critical" {
   threshold           = 0.9 * var.sns_monthly_spend_limit
   treat_missing_data  = "notBreaching"
   alarm_actions       = [aws_sns_topic.notification-canada-ca-alert-critical.arn]
-  ok_actions          = [aws_sns_topic.notification-canada-ca-alert-critical.arn]
+  ok_actions          = [aws_sns_topic.notification-canada-ca-alert-ok.arn]
 }
 
 resource "aws_cloudwatch_metric_alarm" "sns-spending-us-west-2-critical" {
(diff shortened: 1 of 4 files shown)


aws/common/cloudwatch_alarms.tf
```diff
@@ -741,7 +741,7 @@ resource "aws_cloudwatch_metric_alarm" "inflights-not-being-processed-critical"
 }
 
 resource "aws_cloudwatch_metric_alarm" "bulk-not-being-processed-warning" {
-  alarm_name          = "bulk-not-being-processed-warning"
+  alarm_name          = "bulk-buffer-not-being-processed-warning"
```
jimleroyer, line 744: Thanks!

env/production/common/terragrunt.hcl
```diff
@@ -12,8 +12,8 @@ inputs = {
...
   sns_monthly_spend_limit_us_west_2                         = 2000
   alarm_warning_document_download_bucket_size_gb            = 100
-  alarm_warning_inflight_processed_created_delta_threshold  = 50
-  alarm_critical_inflight_processed_created_delta_threshold = 100
-  alarm_warning_bulk_processed_created_delta_threshold      = 50
-  alarm_critical_bulk_processed_created_delta_threshold     = 100
+  alarm_warning_inflight_processed_created_delta_threshold  = 100
+  alarm_critical_inflight_processed_created_delta_threshold = 200
+  alarm_warning_bulk_processed_created_delta_threshold      = 5000
+  alarm_critical_bulk_processed_created_delta_threshold     = 10000
```
jimleroyer, line 18: That escalated quickly. 😅

aws/common/sns.tf
```diff
@@ -119,3 +157,15 @@ resource "aws_sns_topic_subscription" "alert_critical_us_west_2_to_opsgenie" {
   raw_message_delivery   = false
   endpoint_auto_confirms = true
 }
+
+resource "aws_sns_topic_subscription" "alert_critical_us_west_2_to_opsgenie_ok" {
+  provider = aws.us-west-2
+
+  count = var.env == "production" ? 1 : 0
```
jimleroyer, line 164: How is the count used here?

jimleroyer (approved):
LGTM

aws/common/sns.tf
mohamed-cds: @jimleroyer it checks whether the environment is `production` if it's false it'll set the count to 0 which will result in the resource not being created. It's a common way in TF to only deploy certain resources to prod vs staging, etc.

PR #411, opened 2022-03-09
fix: set OpsGenie and Slack variables as senstive

patheard:
# Summary
This will prevent the variable values from appearing in the GitHub
workflow logs.

# ⚠️  Note
The Slack webhook has been regenerated and updated in the:
* AWS console Lambda configuration,
* GitHub secret, and
* LastPast note.

The reason for manually updating the webhook using the AWS console
was to prevent a gap in webhook regeneration and notifications to
Slack while we waited for the `terraform apply`.


jimleroyer (approved):
LGTM

PR #414, opened 2022-03-15
chore: remove unused workflow steps

mohdnr:
Notify moved away from selective terraform applys, so these steps are no longer necessary since theres no need to identify which files changed.

Minor additional changes:

1. Prevents new versions being tagged when the apply workflow is dispatched manually. This is because manual dispatch will be done by the notify pr-bot to sync environments.
2. Update `aws_secretsmanager_secret_version` name to reference `current`

Related: 

- Production: cds-snc/notification-terraform/pull/61
- Staging: cds-snc/notification-terraform/pull/239

### DIFF
--- a/aws/common/outputs.tf
+++ b/aws/common/outputs.tf
@@ -60,6 +60,6 @@ output "s3_bucket_csv_upload_bucket_name" {
 
 output "environment_variables_current_secret_string" {
   description = "Environment variables shared between EKS and Lambda"
-  value       = data.aws_secretsmanager_secret_version.environment_variables_current.secret_string
+  value       = data.aws_secretsmanager_secret_version.current.secret_string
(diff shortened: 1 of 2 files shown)


.github/workflows/merge_to_main_staging.yml
```diff
@@ -144,6 +128,7 @@ jobs:
           ../../../bin/terragrunt apply --terragrunt-non-interactive -auto-approve
 
       - name: Bump version and push tag
+        if: github.event_name != 'workflow_dispatch' # We don't want to tag new versions when launched via workflow_dispatch since only environment variables changed
```
jimleroyer, line 131: Good comment!

jimleroyer (approved):

PR #417, opened 2022-03-16
feat: release v0.115.0

mohamed-cds:
# Summary

- This creates the secretsmanager resource that will be used to share secrets between the k8 and lambda environments.
- Markers webhooks as sensitive
- Workflow cleanup

❗**Note**: The apply will fail one time when terraform attempts to retrieve the latest secrets. Once the manifest repo sync is run, this will no longer happen in future terraform apply runs.

# Release diff
https://github.com/cds-snc/notification-terraform/compare/v0.113.1...v0.115.0


jimleroyer:
@mohamed-cds We should probably have @sastels perform the release as he is the ops lead this week.

PR #421, opened 2022-03-21
Revert lambda/k8s environment sync due to size limit errors

mohamed-cds:
Revert changes to lambda environment until 4kb limit problem is resolved.


patheard (commented):
Makes sense.  Do you also want to remove the common `environment_variables_current_secret_string ` output since it's not used anymore?

https://github.com/cds-snc/notification-terraform/blob/main/aws/common/outputs.tf#L61-L65

patheard (approved):

PR #423, opened 2022-03-22
feat: policy to allow retrieval of environment param

patheard:
# Summary
Allow the Lambda API to retrieve environment variables from the parameter store.

# Related
* cds-snc/notification-planning#421


mohamed-cds (approved):
🚀 ✅

PR #425, opened 2022-03-22
chore: remove envs provided upstream in the manifest

mohamed-cds:
Lambda envs that are common with the manifest repo are no longer necessary since the api will automatically retrieve the common envs from aws system manager parameter store.

### DIFF
--- a/aws/lambda-api/lambda.tf
+++ b/aws/lambda-api/lambda.tf
@@ -25,6 +25,7 @@ resource "aws_lambda_function" "api" {
       ADMIN_CLIENT_SECRET                   = var.admin_client_secret
       DOCUMENT_DOWNLOAD_API_HOST            = var.document_download_api_host
       SQLALCHEMY_DATABASE_URI               = var.sqlalchemy_database_uri
+      SQLALCHEMY_DATABASE_READER_URI        = var.sqlalchemy_database_reader_uri
       NOTIFICATION_QUEUE_PREFIX             = var.notification_queue_prefix
       NOTIFY_EMAIL_DOMAIN                   = var.domain
       NOTIFY_ENVIRONMENT                    = var.env
--- a/aws/lambda-api/lambda.tf
+++ b/aws/lambda-api/lambda.tf
@@ -44,6 +44,7 @@ resource "aws_lambda_function" "api" {
   lifecycle {
     ignore_changes = [
       image_uri,
+      description, # Will be updated outside TF to force cold start existing lambdas. Primarily when common envs are updated
     ]
   }
 }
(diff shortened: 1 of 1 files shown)


aws/lambda-api/lambda.tf
```diff
@@ -23,16 +23,13 @@ resource "aws_lambda_function" "api" {
...
-      DOCUMENTS_BUCKET                      = var.documents_bucket
-      ENVIRONMENT                           = var.env
-      MLWR_HOST                             = var.mlwr_host
+      DOCUMENT_DOWNLOAD_API_HOST            = var.document_download_api_host
+      SQLALCHEMY_DATABASE_URI               = var.sqlalchemy_database_uri
+      SQLALCHEMY_DATABASE_READER_URI        = var.sqlalchemy_database_reader_uri
+      NOTIFICATION_QUEUE_PREFIX             = var.notification_queue_prefix
+      NOTIFY_EMAIL_DOMAIN                   = var.domain
+      NOTIFY_ENVIRONMENT                    = var.env
+      REDIS_ENABLED                         = var.redis_enabled
```
patheard, line 32: I think these 4 can be removed since they should be the same as what we have in the encrypted `.env`.

patheard, line 32: I think these 4 can be removed since they should be the same as what we have in the encrypted `.env`.

aws/lambda-api/lambda.tf
```diff
@@ -41,26 +38,14 @@ resource "aws_lambda_function" "api" {
...
-      REDIS_URL                             = var.redis_url
-      SECRET_KEY                            = var.secret_key
-      SQLALCHEMY_DATABASE_READER_URI        = var.sqlalchemy_database_reader_uri
-      SQLALCHEMY_DATABASE_URI               = var.sqlalchemy_database_uri
-      SQLALCHEMY_POOL_SIZE                  = var.sqlalchemy_pool_size
-      DOCUMENT_DOWNLOAD_API_HOST            = var.document_download_api_host
-      FF_BATCH_INSERTION                    = var.ff_batch_insertion
-      FF_REDIS_BATCH_SAVING                 = var.ff_redis_batch_saving
       FF_CLOUDWATCH_METRICS_ENABLED         = var.ff_cloudwatch_metrics_enabled
-      FF_NOTIFICATION_CELERY_PERSISTENCE    = var.ff_notification_celery_persistence
```
patheard, line 57: Same deal about non-impacted lines.  I'd say `FF_CLOUDWATCH_METRICS_ENABLED` can be deleted as well.

patheard: This is being hardcoded in the k8s manifests.

aws/lambda-api/lambda.tf
patheard: All 4 of these are hardcoded or massaged from different named env vars in the k8s manifests.

patheard (approved):

PR #429, opened 2022-03-25
feat: VPC, WAF and LB logs to Cloud Based Sensor

patheard:
# Summary
Update the Notify infrastructure to send VPC flow logs, WAF ACL logs
and LB access logs to the Cloud Based Sensor satellite bucket.

# Related
* cds-snc/cloud-based-sensor#34
* cds-snc/notification-planning#525
* cds-snc/notification-planning#531

### DIFF
--- a/aws/common/athena.tf
+++ b/aws/common/athena.tf
@@ -40,6 +40,6 @@ resource "aws_athena_named_query" "create_table_alb_logs" {
     {
       database_name   = aws_athena_database.notification_athena.name
       table_name      = "alb_logs"
-      bucket_location = "s3://${aws_s3_bucket.alb_log_bucket.bucket}/AWSLogs/${var.account_id}/elasticloadbalancing/${var.region}"
+      bucket_location = "s3://${var.cbs_satellite_bucket_name}/lb_logs/AWSLogs/${var.account_id}/elasticloadbalancing/${var.region}"
   })
 }
--- a/aws/eks/waf.tf
+++ b/aws/eks/waf.tf
@@ -274,11 +274,6 @@ resource "aws_wafv2_web_acl_logging_configuration" "cloudwatch-waf-logs" {
 #
 # WAF logging to Cloud Based Sensor satellite bucket
 #
-locals {
-  cbs_satellite_bucket_arn    = "arn:aws:s3:::${var.cbs_satellite_bucket_name}"
-  cbs_satellite_bucket_prefix = "waf_acl_logs/AWSLogs/${var.account_id}/"
-}
-
 resource "aws_kinesis_firehose_delivery_stream" "firehose-waf-logs" {
   name        = "aws-waf-logs-notification-canada-ca-waf"
   destination = "extended_s3"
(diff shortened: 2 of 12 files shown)


aws/eks/waf.tf
```diff
@@ -260,3 +260,86 @@ resource "aws_cloudwatch_log_group" "notification-canada-ca-waf-logs" {
...
+
+resource "aws_kinesis_firehose_delivery_stream" "firehose-waf-logs" {
+  name        = "aws-waf-logs-notification-canada-ca-waf"
+  destination = "extended_s3"
+
+  server_side_encryption {
+    enabled = true
+  }
+
+  extended_s3_configuration {
```
mohamed-cds, line 280: should we increase the buffer_size ? Default is 5MB. AWS recommends setting buffer_size in MBs to a value greater than the amount of data you typically ingest into the delivery stream in 10 seconds.

Notify receives alot of traffic and was curious if there's any potential downstream impacts of keeping this at the default 5MB.

patheard: Ah, that's a good point about the default `buffer_size`.  

I checked out prod's `IncomingBytes` metric for the `aws-waf-logs-notification-canada-ca-waf` CloudWatch log group and the max I'm seeing historically for a 5 minute period is 161kb.  With a 10 second period I've only got 3 hours of data, but the max I see is 20kb.  I'll check again today after the HC send to see what it looks like.

mohamed-cds (approved):
LGTM 🚀 ✅

aws/eks/waf.tf
```diff
@@ -347,3 +347,13 @@ data "aws_iam_policy_document" "firehose-waf-logs" {
...
   }
 }
+
+resource "aws_wafv2_web_acl_logging_configuration" "firehose-waf-logs" {
+  log_destination_configs = [aws_kinesis_firehose_delivery_stream.firehose-waf-logs.arn]
+  resource_arn            = aws_wafv2_web_acl.notification-canada-ca.arn
+  redacted_fields {
+    single_header {
+      name = "authorization"
+    }
```
mohamed-cds, line 357: nice!

patheard: No need - weirdly the `aws_wafv2_web_acl_logging_configuration` wants you to specify [everything as lowercase](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl_logging_configuration#single-header), so my assumption is that this conversion happens automagically.

aws/common/athena.tf
```diff
@@ -40,6 +40,6 @@ resource "aws_athena_named_query" "create_table_alb_logs" {
     {
       database_name   = aws_athena_database.notification_athena.name
       table_name      = "alb_logs"
-      bucket_location = "s3://${aws_s3_bucket.alb_log_bucket.bucket}/AWSLogs/${var.account_id}/elasticloadbalancing/${var.region}"
+      bucket_location = "s3://${var.cbs_satellite_bucket_name}/lb_logs/AWSLogs/${var.account_id}/elasticloadbalancing/${var.region}"
```
jimleroyer, line 43: Thanks for switching that one too, although I am not sure it is functional. 

Will the new S3 bucket be available to CBS for access?

patheard: I tested before with the old ALB log bucket and it appeared to work, but I have the same question as you since I didn't find any IAM policies granting Athena permission to pull from the old ALB bucket.

My plan was to test in Staging once this merged to see if Athena was able to pull data from the new CBS S3 bucket.

PR #431, opened 2022-03-30
chore: remove unnecessary newrelic envs

mohamed-cds:
These two environment variables are handled by the common parameter store env sync and are incorrectly overriding. Removing them will make the lambda fallback to the common variable retrieved from the parameter store.


patheard (changes requested):
Nice!  Could you also remove the `TF_VAR_new_relic_license_key` from the workflows?

patheard (approved):

PR #434, opened 2022-03-31
feat: create Firehose WAF logging IAM role

patheard:
# Summary
This role will be used by a Kinesis Firehose to send WAF ACL
logs to the Cloud Based Sensor satellite bucket.

It is being created first so that its ARN can be used as an
output by the EKS and Lambda API Terraform modules.

# Related
*  #429 
* cds-snc/notification-planning#525


mohamed-cds (approved):
LGTM 🚀 ✅

PR #438, opened 2022-04-06
fix: remove CloudWatch WAF ACL log destination

patheard:
# Summary
Delete the CloudWatch WAF ACL logging destination as the S3
bucket + Firehose is required by Cloud Based Sensor. An Athena
query has been added to create a table from the WAF logs.

Also fixes the Athena query that create's a table from the load balancer
S3 access logs bucket. The previous query was running successfully, 
but failing to pull any data into the new table.

### DIFF
--- a/aws/common/athena.tf
+++ b/aws/common/athena.tf
@@ -52,6 +52,6 @@ resource "aws_athena_named_query" "create_table_waf_logs" {
     {
       database_name   = aws_athena_database.notification_athena.name
       table_name      = "waf_logs"
-      bucket_location = "s3://${var.cbs_satellite_bucket_name}/waf_acl_logs/AWSLogs/${var.account_id}/lb/${var.region}/"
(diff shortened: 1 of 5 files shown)


sastels (approved):
LGTM. 
We don't have any metrics or alerts based on those log groups?

patheard:
> LGTM. We don't have any metrics or alerts based on those log groups?

No, we added it a while back since we didn't have an WAF ACL logging configured and wanted to test out using CloudWatch, which had just recently been added as an option.

jimleroyer:
So we removed cloudwatch because it's either the s3 bucket+firehose or cloudwatch? This is handy to have logs in cloudwatch.

patheard:
Yeah, sadly a WAF ACL can only support one logging destination 😢 

To try and make up for it, I added a new saved query that can create an Athena table from the WAF logs in the S3 bucket.

PR #447, opened 2022-04-19
chore: set Lambda API min concurrency to 3

patheard:
# Summary 
Update the Lambda API's provisioned minimum concurrency during
high demand periods (12pm - 5pm).  This is because we're now serving
100% of our API requests through the Lambda API.

# ⚠️  Note
There should be no infra changes in Staging for this PR.

# Related
* cds-snc/notification-planning#438


sastels (approved):
LGTM. Can we monitor how many lambdas get spun up to keep an eye on when we might want to increase the min and max here?

patheard:
> LGTM. Can we monitor how many lambdas get spun up to keep an eye on when we might want to increase the min and max here?

Indeed there is - we get rough metrics through the [Lambda API's Monitoring dashboard](https://ca-central-1.console.aws.amazon.com/lambda/home?region=ca-central-1#/functions/api-lambda?tab=monitoring), which shows us fluctuating between 10 - 20 concurrent executions during our peak time.

I'm going to look into [enabling Lambda Insights](https://docs.aws.amazon.com/lambda/latest/dg/monitoring-insights.html) to give us better metrics around how the Lambda is performing.

PR #448, opened 2022-04-19
feat: workflow to check if k8s cluster upgrades

patheard:
# Summary
Weekly GitHub workflow that will post to Slack if there is a
version upgrade available for the Kubernetes cluster.

Also updates the AMI version workflow to use the cluster's
current version to check for available image updates.


jimleroyer (approved):
LGTM

PR #452, opened 2022-04-26
feat: k8s v1.21 upgrade and EKS addons

patheard:
# Summary
Upgrade the Kubernetes cluster to v1.21 and bump node
worker AMI release versions.

Also adds Terraform resources to manage EKS CoreDNS,
kube-proxy and vpc-cni addon versions.

# Related

### DIFF
--- a/env/production/eks/terragrunt.hcl
+++ b/env/production/eks/terragrunt.hcl
@@ -39,5 +39,8 @@ inputs = {
   cloudfront_assets_arn                  = dependency.cloudfront.outputs.cloudfront_assets_arn
   eks_cluster_name                       = "notification-canada-ca-production-eks-cluster"
   eks_cluster_version                    = "1.20"
+  eks_addon_coredns_version              = "v1.8.3-eksbuild.1"
+  eks_addon_kube_proxy_version           = "v1.20.4-eksbuild.2"
+  eks_addon_vpc_cni_version              = "v1.11.0-eksbuild.1"  
   eks_node_ami_version                   = "1.20.11-20220406"
 }
--- a/env/staging/eks/terragrunt.hcl
+++ b/env/staging/eks/terragrunt.hcl
@@ -68,8 +68,11 @@ inputs = {
   firehose_waf_logs_iam_role_arn         = dependency.common.outputs.firehose_waf_logs_iam_role_arn
   cloudfront_assets_arn                  = dependency.cloudfront.outputs.cloudfront_assets_arn
   eks_cluster_name                       = "notification-canada-ca-staging-eks-cluster"
-  eks_cluster_version                    = "1.20"
-  eks_node_ami_version                   = "1.20.11-20220406"  
+  eks_cluster_version                    = "1.21"
+  eks_addon_coredns_version              = "v1.8.3-eksbuild.1"
+  eks_addon_kube_proxy_version           = "v1.20.4-eksbuild.2"
+  eks_addon_vpc_cni_version              = "v1.11.0-eksbuild.1"
+  eks_node_ami_version                   = "1.21.5-20220420"
 }
 
 terraform {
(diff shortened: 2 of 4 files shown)


env/production/eks/terragrunt.hcl
```diff
@@ -39,5 +39,8 @@ inputs = {
   cloudfront_assets_arn                  = dependency.cloudfront.outputs.cloudfront_assets_arn
   eks_cluster_name                       = "notification-canada-ca-production-eks-cluster"
   eks_cluster_version                    = "1.20"
```
sastels, line 41: change this too?

env/production/eks/terragrunt.hcl
```diff
@@ -39,5 +39,8 @@ inputs = {
...
+  eks_addon_coredns_version              = "v1.8.3-eksbuild.1"
+  eks_addon_kube_proxy_version           = "v1.20.4-eksbuild.2"
+  eks_addon_vpc_cni_version              = "v1.11.0-eksbuild.1"  
   eks_node_ami_version                   = "1.20.11-20220406"
```
sastels, line 45: and this?

env/staging/eks/terragrunt.hcl
```diff
@@ -68,8 +68,11 @@ inputs = {
   firehose_waf_logs_iam_role_arn         = dependency.common.outputs.firehose_waf_logs_iam_role_arn
   cloudfront_assets_arn                  = dependency.cloudfront.outputs.cloudfront_assets_arn
   eks_cluster_name                       = "notification-canada-ca-staging-eks-cluster"
-  eks_cluster_version                    = "1.20"
-  eks_node_ami_version                   = "1.20.11-20220406"  
+  eks_cluster_version                    = "1.21"
+  eks_addon_coredns_version              = "v1.8.3-eksbuild.1"
```
sastels, line 72: Should this be 1.8.4? looking at https://docs.aws.amazon.com/eks/latest/userguide/managing-coredns.html

env/staging/eks/terragrunt.hcl
```diff
@@ -68,8 +68,11 @@ inputs = {
...
+  eks_addon_kube_proxy_version           = "v1.20.4-eksbuild.2"
+  eks_addon_vpc_cni_version              = "v1.11.0-eksbuild.1"
+  eks_node_ami_version                   = "1.21.5-20220420"
```
sastels, line 75: oh, are these the versions for k8s 1.20, and after the 1.21 update we'd roll these ahead?

env/production/eks/terragrunt.hcl
patheard: I want to leave these as `1.20` until we're ready to do the prod upgrade.  Reason being is that our `terraform apply` is a bit odd in that it uses the `infrastructure_version` tag to determine which tag of `aws` modules to plan/apply, but takes the `terragrunt.hcl` files from `main`.

As a result, if I bump this now, and then a different TF change goes to prod before we're expecting the upgrade, we'll still get the upgrade 😅

env/production/eks/terragrunt.hcl
patheard: Same reason as leaving the `eks_cluster_version=1.20`

env/staging/eks/terragrunt.hcl
patheard: It should be, but I'd like to do that EKS addon bump in a separate PR just do double check my TF state import worked like I expected.  In theory though, you're totally right - it sould be `1.8.4` for k8s 1.21.

env/staging/eks/terragrunt.hcl
patheard: Yup, that's exactly it.

sastels (approved):
ok, I remember how this all works now. This is awesome, It'll make it so much easier than the manual process!

PR #453, opened 2022-04-26
chore: upgrade coredns/kube-proxy EKS addons

patheard:
# Summary
Upgrade the CoreDNS and kube-proxy EKS addons to the
recommended versions for Kubernetes 1.21.

# Related


sastels (approved):
Thanks for linking to the docs!
I see from the docs that the cni is the same for 1.21 and 1.20 👍 
🚀

PR #458, opened 2022-05-03
Add new database queues and alarms

sastels:
# Summary | Résumé

- Add alarms for new priority based database task queues
- explicitly add the database queues and make the high priority queue and a FIFO (https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/sqs_queue#high-throughput-fifo-queue)


sastels:
I see that none of our other queues are encrypted - should we just ignore this error, or actually turn on encryption? 

Vaguely related: it appears that the queues are automatically created by Notify (possibly when it first posts to a queue?) - should these be defined in terraform instead?

sastels:
Will have to ignore that warning anyway, even with AWS managed encryption:
https://deepsource.io/directory/analyzers/terraform/issues/TF-AWS015
> Unencrypted SQS queueTF-AWS015
Queues should be encrypted with customer managed KMS keys and not default AWS managed keys, in order to allow granular control over access to specific queues.

aws/common/variables.tf
```diff
@@ -75,6 +75,27 @@ variable "sqs_db_tasks_queue_name" {
   default = "database-tasks"
 }
 
+variable "sqs_priority_db_tasks_queue_name" {
+  type = string
+  # See QueueNames in
+  # https://github.com/cds-snc/notification-api/blob/master/app/config.py
+  default = "priority-database-tasks.fifo"
```
patheard, line 82: A minor preference thing - could you move these values out into the `terragrunt.hcl` for each environment rather than relying on the `default` variable values?  I find it tricky to track down values that are coming from defaults.

aws/common/cloudwatch_alarms.tf
```diff
@@ -478,6 +478,106 @@ resource "aws_cloudwatch_metric_alarm" "sqs-db-tasks-stuck-in-queue-critical" {
...
+  alarm_name          = "sqs-normal-db-tasks-stuck-in-queue-critical"
+  alarm_description   = "ApproximateAgeOfOldestMessage in normal DB tasks queue is older than 15 minute for 1 minute"
+  comparison_operator = "GreaterThanOrEqualToThreshold"
+  evaluation_periods  = "15"
+  metric_name         = "ApproximateAgeOfOldestMessage"
+  namespace           = "AWS/SQS"
+  period              = 60
+  statistic           = "Maximum"
+  threshold           = 60 * 15
+  alarm_actions       = [aws_sns_topic.notification-canada-ca-alert-critical.arn]
```
patheard, line 540: Question - do we want to tune the thresholds before we start triggering critical alarms and OpsGenie pages?

sastels: I just copied the current thresholds for our existing database task (that we're splitting into these three queues), so it shouldn't be an issue. Possibly we should lower some of these though, in particular the one for priority tasks. I would say we first run the system and monitor for a while to get a sense of what's normal for that queue.

env/production/common/terragrunt.hcl
```diff
@@ -16,4 +16,10 @@ inputs = {
   alarm_critical_inflight_processed_created_delta_threshold = 200
   alarm_warning_bulk_processed_created_delta_threshold      = 5000
   alarm_critical_bulk_processed_created_delta_threshold     = 10000
+  sqs_priority_db_tasks_queue_name                          = "priority-database-tasks.fifo"
+  sqs_normal_db_tasks_queue_name                            = "normal-database-tasks"
+  sqs_bulk_db_tasks_queue_name                              = "bulk-database-tasks"
```
patheard, line 21: Thank you!

patheard (approved):
Looks good!  I had one minor comment about queue names, but it's definitely a preference thing.

aws/common/sqs.tf
```diff
@@ -0,0 +1,23 @@
+resource "aws_sqs_queue" "priority_db_tasks_queue" {
+  name                    = "${var.celery_queue_prefix}${var.sqs_priority_db_tasks_queue_name}"
```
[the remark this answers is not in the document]
sastels: It's a bit ugly, we have for celery (in api) https://github.com/cds-snc/notification-api/blob/main/app/config.py#L253-L258
```
 BROKER_TRANSPORT_OPTIONS = {
        "region": AWS_REGION,
        "polling_interval": 1,  # 1 second
        "visibility_timeout": 310,
        "queue_name_prefix": NOTIFICATION_QUEUE_PREFIX,
    }
```

where that `NOTIFICATION_QUEUE_PREFIX` is `eks-notification-canada-ca` and in api we will define stuff like
```
QueueName.PRIORITY_DATABASE = "priority-database-tasks.fifo"
```
so that gets put together to "eks-notification-canada-capriority-database-tasks.fifo"

We can add a "-" to the start of the QueueNames though, which looks ugly in python, but we always will reference it as `QueueName.PRIORITY_DATABASE` so doesn't affect the python dev experience there.

And it'll make the AWS experience better, admittedly 👍

PR #459, opened 2022-05-03
fix: update IAM policy for ALB controller

patheard:
# Summary
This is required for the AWS load balancer 
controller v2.4.1 upgrade.

For reference, the AWS load balancer controller pods
are being provided access to this IAM policy through
inheritance via the EKS worker node.

# Related
* cds-snc/notification-planning#643

### DIFF
--- a/aws/eks/alb_iam_policy.json
+++ b/aws/eks/alb_iam_policy.json
@@ -4,16 +4,31 @@
         {
             "Effect": "Allow",
             "Action": [
-                "iam:CreateServiceLinkedRole",
+                "iam:CreateServiceLinkedRole"
+            ],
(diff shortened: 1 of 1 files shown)


jimleroyer (approved):
It's the upgrade that requires these new permissions?

iokpala (approved):
LGTM

patheard:
> It's the upgrade that requires these new permissions?

It's the [AWS Load Balancer Controller](https://docs.aws.amazon.com/eks/latest/userguide/aws-load-balancer-controller.html) upgrade that needs these permissions, and we need to do this upgrade before going to Kubernetes 1.22.

PR #461, opened 2022-05-04
fix: make legacy S3 asset bucket private

patheard:
# Summary
This bucket is no longer used and has been replaced by the
`notification-canada-ca-production-asset-upload` S3 bucket.

As a first step, we'll make the bucket private and then check if
it can be fully removed.

# ⚠️  Note
This will have no `terraform plan` changes for Staging as it's
only created in Production.


CalvinRodo (approved):
from an infra as code point of view this looks good.

PR #466, opened 2022-05-06
fix: do not publish new Lambda versions

patheard:
# Summary
By default, Terraform does not publish new Lambda versions
when a create/change operation is performed.

This is the behaviour we would like as it will allow our
provisioned concurrency to stay in sync with the Lambda
function updates performed by the AWS cli in our
release workflows.


sastels (approved):
🚀

PR #467, opened 2022-05-06
fix: add Lamda API function alias

patheard:
# Summary
Creates an alias for the Lambda API that always references the
`$LATEST` deployed version of the function.

This alias is then used by the provisioned concurrency config to
always keep the execution environments in sync with the latest
deployed changes.


jimleroyer (approved):
LGTM

PR #468, opened 2022-05-06
fix: update lambda alias to reference version

patheard:
# Summary
Provisioned concurrency cannot be used with an unpublished
lambda function.  This updates the lambda to publish new
versions on changes and update the alias with that new version.


jimleroyer (approved):

sastels (approved):
LGTM

PR #471, opened 2022-05-09
release: v0.127.9 Lambda alias and IAM updates

patheard:
# Summary
Update Lambda API to use an alias for provisioned concurrency.
Also updates the IAM permissions for the `ecr-user` so that it
can update the Lambda's alias and publish a new function version.

# Diff
https://github.com/cds-snc/notification-terraform/compare/v0.127.6...v0.127.9


jimleroyer (approved):
:shipit:

PR #473, opened 2022-05-09
release: v0.127.10 Lambda API autoscaling target

patheard:
# Summary
Update the Lambda API autoscaling target to use the alias
rather than the Lambda version.

# Diff
https://github.com/cds-snc/notification-terraform/compare/v0.127.9...v0.127.10


jimleroyer (approved):
:shipit:

PR #475, opened 2022-05-09
release: v0.127.12 Kubernetes v1.22 upgrade

patheard:
# Summary
Upgrade Kubernetes to v1.22 and update the node
AMI release version.

# Diff
https://github.com/cds-snc/notification-terraform/compare/v0.127.10...v0.127.12

# Related


.github/workflows/merge_to_main_production.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           echo "INFRASTRUCTURE_VERSION=$INFRASTRUCTURE_VERSION" >> $GITHUB_ENV
 
       - name: Setup Terraform
-        uses: hashicorp/setup-terraform@v1.2.1
+        uses: hashicorp/setup-terraform@17d4c9b8043b238f6f35641cdd8433da1e6f3867 # v2.0.0
```
jimleroyer, line 61: The tag isn't created yet?

jimleroyer (approved):

.github/workflows/merge_to_main_production.yml
patheard: So this was me testing an odd failure with the `hashicorp/setup-terraform` plan to see if the `v2.0.0` of the action worked.  I suspect it was failing because of the Hashicorp scheduled maintenance.

That being said, it made me realize we can switch over to using CDS's Terraform setup action to get rid of another workflow dependency.

patheard: Actually, thinking more about it, I've reverted all workflow changes and will do this in another PR to limit the changes around this PR.

PR #478, opened 2022-05-10
fix: update API gateway to invoke Lambda alias

patheard:
# Summary

The Lambda API is setup with provisioned concurrency attached
to the `latest` Lambda alias.  By using this alias with the API
gateway integration, we'll be able to take advantage of that
provisioned concurrency.

Currently the API gateway is always invoking the `$LATEST` version
of the function code, which means the provisioned concurrency
is not used.

# Note
This PR also updates the `terraform plan/apply` workflows to use
the [CDS Terraform tools setup action](https://github.com/cds-snc/terraform-tools-setup) to remove the
`hashicorp/setup-terraform` action dependency.

# Related
* cds-snc/notification-planning#646


jimleroyer (approved):
LGTM and aligned on changes discussed yesterday in he call.

PR #481, opened 2022-05-18
fix: redeploy API gateway when config changes

patheard:
# Summary
This fix will cause the API gateway to be redeployed when there
are any configuration changes made.

Also updates the Lambda permissions to allow the API gateway
to execute the `latest` Lambda alias.


jimleroyer (approved):
:shipit:

PR #482, opened 2022-05-19
feat: Lambda Admin PR review environment

patheard:
# Summary
Add the ECR and IAM role used by the Notify Admin pull request
review environment.  These are only needed in the Notify Staging account.

# ⚠️  Note
This resources were initially created in the AWS console, so not all
of them will be shown in the Terraform plan.


jimleroyer (approved):
Thanks @patheard for closing the loop.

PR #488, opened 2022-06-27
tweak Cloudwatch Insights query

sastels:
# Summary | Résumé

We have Notify errors that don't have levelname ERROR, such as
```
{"name": "api", "levelname": "INFO", "message": "{'result': 'error', 'message': 'Template not found'}",
```
 
Run and compare both versions of the query.


jimleroyer (approved):

patheard (approved):
🧹

PR #491, opened 2022-07-11
Release 0.128.5

jimleroyer:
# Summary | Résumé

* [Lowering data retention period for CSV uploads](https://github.com/cds-snc/notification-terraform/pull/490)

* Check the S3 bucket once this comes in effect to see if there are documents still in the bucket that are older than 7 days.


sastels (approved):
plan looks good!

PR #493, opened 2022-07-18
Added api sub-domain in the list of alt names in certificates

jimleroyer:
# Summary | Résumé

Trying out to add a subdomain for `*.api.notification.canada.ca` as an alt name for our certificates, helping our friends using WebLogic as an app container, as indicated by [this Freshdesk ticket](https://cds-snc.freshdesk.com/a/tickets/9116).

Also removed an `ignore_changes` statement on a bug previously identified but that should not take place anymore with hashicorp/aws/3.0.0. Let's see how that works out.

1. Head to staging or production environment with the new deployed certificate (e.g. `api.notification.canada.ca`).
2. Get the certificate details.
3. Check the Subject Alternative Name field and check if there is an entry for the new `*.api.notification.canada.ca` entry.

Here is a screenshot of the current certificate details without the new entry for the API sub-domain:

image

### DIFF
--- a/aws/dns/acm.tf
+++ b/aws/dns/acm.tf
@@ -22,7 +22,7 @@ resource "aws_acm_certificate" "notification-canada-ca-alt" {
   domain_name = var.alt_domain
   subject_alternative_names = [
     "*.${var.alt_domain}",
-    "*.api.${var.domain}",
+    "*.api.${var.alt_domain}",
     "*.document.${var.alt_domain}"
   ]
   validation_method = "DNS"
--- a/aws/dns/acm.tf
+++ b/aws/dns/acm.tf
@@ -9,9 +9,6 @@ resource "aws_acm_certificate" "notification-canada-ca" {
 
   lifecycle {
     create_before_destroy = true
-    # TF bug on AWS 2.0: prevents certificates from being destroyed/recreated
-    # https://github.com/hashicorp/terraform-provider-aws/issues/8531
-    ignore_changes = [subject_alternative_names]
   }
 
   tags = {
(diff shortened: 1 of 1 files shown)


aws/dns/acm.tf
```diff
@@ -2,15 +2,13 @@ resource "aws_acm_certificate" "notification-canada-ca" {
...
+    "*.api.${var.domain}",
     "*.document.${var.domain}"
   ]
   validation_method = "DNS"
 
   lifecycle {
     create_before_destroy = true
-    # TF bug on AWS 2.0: prevents certificates from being destroyed/recreated
-    # https://github.com/hashicorp/terraform-provider-aws/issues/8531
-    ignore_changes = [subject_alternative_names]
```
jimleroyer, line 13: Based on the [provided linked issue](https://github.com/hashicorp/terraform-provider-aws/issues/8531#issuecomment-663562156), this should have been fixed by version hashicorp/aws/3.0.0 which we switched to 18 months ago. That `ignore_changes` statement was introduced 20 months ago though and we might never updated this (as we didn't modify that yet).

aws/dns/acm.tf
```diff
@@ -24,15 +22,13 @@ resource "aws_acm_certificate" "notification-canada-ca-alt" {
   domain_name = var.alt_domain
   subject_alternative_names = [
     "*.${var.alt_domain}",
+    "*.api.${var.domain}",
```
mohamed-cds, line 25: this block is for the `alpha.canada.ca` domain so i think you need to switch this var to `var.alt_domain`

mohamed-cds (commented):
Minor change suggested but other than that LGTM since it's the same as what i did in the [security-tools project](https://github.com/cds-snc/security-tools/blob/main/terragrunt/aws/sso_proxy/acm.tf#L3-L6)

aws/dns/acm.tf
jimleroyer: oh right!

mohamed-cds (approved):

sastels (approved):
Let's see if it all works!

PR #494, opened 2022-07-19
tweak WAF regex for base64

sastels:
# Summary | Résumé

`\w` matches word characters, for base64 we need to match any non-whitespace.


jimleroyer (approved):
:shipit:

PR #495, opened 2022-07-19
Revert addition of api sub-domain 

jimleroyer:
# Summary | Résumé

Because the addition of a new domain in the alternative names in our certificates proved more difficult than initially anticipated, I am reverting previously merged change around _addition of api sub-domain in the list of alt names in certificates_ to unblock the TF release pipeline.

The old certificates can't be properly destroyed at the moment as they kept being attached to the load balancer listener. We might have to recreate the load balancer in a way that won't affect downtime in production.

Setup should be as before without plan and TF apply failures.

### DIFF
--- a/aws/dns/acm.tf
+++ b/aws/dns/acm.tf
@@ -2,13 +2,15 @@ resource "aws_acm_certificate" "notification-canada-ca" {
   domain_name = var.domain
   subject_alternative_names = [
     "*.${var.domain}",
-    "*.api.${var.domain}",
     "*.document.${var.domain}"
   ]
   validation_method = "DNS"
 
   lifecycle {
     create_before_destroy = true
+    # TF bug on AWS 2.0: prevents certificates from being destroyed/recreated
+    # https://github.com/hashicorp/terraform-provider-aws/issues/8531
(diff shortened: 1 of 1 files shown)


sastels:
```
# aws_acm_certificate.notification-canada-ca (deposed object 03a87456) will be destroyed
# (left over from a partially-failed replacement of this instance)
```
I guess this is just deleting the new cert, and the old one is already there.

sastels (approved):
LGTM assuming we're just deleting the new cert.

jimleroyer:
Ok I'll wait to merge this and sync with @mohamed-cds tomorrow morning. The currently referenced certificate planned for deletion is the one currently in use (the old certificate), i.e. `b8ba5e08-2c0f-4d1a-afb1-e41accb271bb`. So I wonder if we'll have to run a few TF commands to manually update the state.

mohamed-cds (changes requested):
reverting no longer necessary since we found a path forward. This can be closed. Blocking so it doesn't get accidentally merged.

jimleroyer:
Closing the revert as we found a way forward.

PR #498, opened 2022-07-25
feat: Block traffic matching common rule set 

jimleroyer:
# Summary | Résumé

Set the common rule set to block by removing the `count` override and letting the `block` default take in. The following rules in the set were excluded as we observed legitimate traffic going through these:

* CrossSiteScripting_BODY
* EC2MetaDataSSRF_BODY
* ~GenericRFI_BODY~: After some checks, it does not appear to get any legitimate traffic so that should be blocked too.
* NoUserAgent_HEADER
* SizeRestrictions_BODY

(Steve) I also excluded GenericLFI_Body because I was seeing legit api lambda traffic being COUNTed, See [CommonRules Analysis](https://docs.google.com/spreadsheets/d/1XBbIBXJWq8nTDVXhwwj-puzHpDYVwJAGajuQG3-QBgU/edit#gid=0).

* Run the automated smoke tests.
* Manual smoke tests.


aws/eks/waf.tf
```diff
@@ -35,13 +35,29 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
       managed_rule_group_statement {
         name        = "AWSManagedRulesCommonRuleSet"
         vendor_name = "AWS"
+
+        excluded_rule {
+          name = "CrossSiteScripting_BODY"
+        }
+
+        excluded_rule {
+          name = "EC2MetaDataSSRF_BODY"
```
jimleroyer, line 51: There is some valid traffic that this rule applies on unfortunately. If we'd want to block on it, we'd need to add more constraints such as the origin country, or the user-agent that would match our own API clients.
```
{clientip=REDACTED, country=CA, headers=[{name=Authorization, value=REDACTED}, {name=User-agent, value=NOTIFY-API-NODE-CLIENT/4.8.0}, {name=host, value=api.notification.alpha.canada.ca}, {name=accept, value=application/json}, {name=content-type, value=application/json}, {name=content-length, value=608}, {name=Connection, value=close}], uri=/v2/notifications/email, args=, httpversion=HTTP/1.1, httpmethod=POST, requestid=1-62eac612-3d1f51924d343bef28cdcbf3}
```

aws/eks/waf.tf
```diff
@@ -35,13 +35,29 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+        }
+
+        excluded_rule {
+          name = "NoUserAgent_HEADER"
```
jimleroyer, line 55: Many requests matching this rule could be blocked but it seems we have a few legitimate ones.

aws/eks/waf.tf
```diff
@@ -35,13 +35,29 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+        }
+
+        excluded_rule {
+          name = "SizeRestrictions_BODY"
```
jimleroyer, line 59: We could exclude that rule only for certain endpoints where we do expects a very large body, such as when uploading files. [Documentation over there](https://docs.aws.amazon.com/waf/latest/developerguide/classic-web-acl-size-conditions.html).

aws/eks/waf.tf
patheard: That's not a bad idea - do we expect any legitimate requests to Notify that are outside of Canada?

patheard (dismissed):

aws/eks/waf.tf
jimleroyer: This could be an additional constraint to add, to enable the rule but only for traffic outside of Canada. This was your thinking?

patheard: Yup, exactly - disable `EC2MetaDataSSRF_BODY` for traffic from Canada, but leave it enabled for all other traffic.

jimleroyer: I'll add a TODO comment beside the rule to not forget about it, as we are currently reworking these. I think that would need some more work to apply a label here and in a further rule apply an action.

sastels:
Are we sure about enabling `GenericRFI_BODY`? You probably looked at the same data I am  (were you including the api lambda logs?), just wondering how you determined that it wasn't legit?

Some data I pulled: https://docs.google.com/spreadsheets/d/1d1AEfts-skjaXsqkVrp4W3Hi2Cho2ejpzoX2mWHWnw4/edit#gid=0

patheard:
> Are we sure about enabling `GenericRFI_BODY`? You probably looked at the same data I am (were you including the api lambda logs?), just wondering how you determined that it wasn't legit?
> 
> Some data I pulled: https://docs.google.com/spreadsheets/d/1d1AEfts-skjaXsqkVrp4W3Hi2Cho2ejpzoX2mWHWnw4/edit#gid=0

Want to have a sync on this?  I'm curious about what the request body looks like on those `GenericRFI_BODY` requests you found.

sastels:
whoops, I need new glasses, it was `GenericLFI_Body` that's counting legit api traffic, not `GenericRFI_Body`
See [CommonRules Analysis](https://docs.google.com/spreadsheets/d/1XBbIBXJWq8nTDVXhwwj-puzHpDYVwJAGajuQG3-QBgU/edit#gid=0)

aws/eks/waf.tf
```diff
@@ -35,13 +35,30 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+          name = "EC2MetaDataSSRF_BODY"
+        }
+        excluded_rule {
+          name = "NoUserAgent_HEADER"
+        }
+        excluded_rule {
+          name = "SizeRestrictions_BODY"
+        }
+        excluded_rule {
+          name = "GenericLFI_Body"
```
sastels, line 60: Matched on a bunch of valid requests to api.

patheard:
I just noticed that you'll need to set the override action to `none {}` on the lambda API as well:
https://github.com/cds-snc/notification-terraform/pull/498/files#diff-c2256a3a91fb5b8ecd0e714c74ac644b86f2e35c8bf0cc6d3793d88c7490b071R37-R39

sastels:
> 

Thanks, good catch!

sastels (approved):

PR #499, opened 2022-07-27
feat: Rate limit login uris

sastels:
# Summary | Résumé

If an ip hits `/sign-in`, `/register`, or `/forgot-password` a combined number of 100 times in a 5 minute period then they will be blocked.

[Analysis of recent data](https://docs.google.com/spreadsheets/d/1Na1ybN1WNJzHFiAEkB97neCkpfpCBrfeRYQjfGKa3mI/edit#gid=1269960622)

Note that 100 hits / 5 minute block is the lowest limit we can set.

- hit these endpoints in staging many times each and together until you're blocked.
- hit other endpoints more times and make sure you don't get blocked


sastels:
We should move the actual limit numbers to variables we send in with terragrunt. I'll make that change.

aws/eks/waf.tf
```diff
@@ -140,6 +140,79 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+
+            statement {
+              byte_match_statement {
+                field_to_match {
+                  uri_path {}
+                }
+                positional_constraint = "STARTS_WITH"
+                search_string         = "/sign-in"
+                text_transformation {
+                  type     = "NONE"
```
patheard, line 179: Just a minor tweak - if you change this to `LOWERCASE` we'll catch cases if someone happens to use a path like `/SIGN-IN`.

aws/eks/waf.tf
```diff
@@ -140,6 +140,79 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+                }
+              }
+            }
+            statement {
+              byte_match_statement {
+                field_to_match {
+                  uri_path {}
+                }
+                positional_constraint = "STARTS_WITH"
+                search_string         = "/forgot-password"
```
jimleroyer, line 203: Do you think it would be useful to have a match for `/forced-password-reset` as well?

jimleroyer (approved):
Looks good to me! I left a comment on the reset password URL too, if we want to add it too.

aws/eks/waf.tf
sastels: yes, good catch 👍

aws/eks/variables.tf
```diff
@@ -83,6 +83,11 @@ variable "eks_node_ami_version" {
   type        = string
 }
 
+variable "sign_in_waf_rate_limit" {
+  description = "Rate limit for /register, /sigh-in, and /forgot-password"
```
jimleroyer, line 87: I just noticed that we only rate limit sighing in, but not signing in. 😉 

(just a typo in the description so nothing bad)

PR #500, opened 2022-07-28
Added a non-wildcard version of the API SAN

jimleroyer:
# Summary | Résumé

Trying out a non-wildcard of the API SAN in addition to a wildcard version. This is to help out the SSC users that are using WebLogic, which can't accept wildcard for the SANs.

Look out for the non-wildcard API entry in the certificate subject alternative names when deployed.

### DIFF
--- a/.github/workflows/terraform_static_analysis.yml
+++ b/.github/workflows/terraform_static_analysis.yml
@@ -18,6 +18,6 @@ jobs:
       - name: Terraform security scan
         uses: triat/terraform-security-scan@v2.2.1
         with:
-          tfsec_version: 'v1.26.3'
+          tfsec_version: 'v0.39.26'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


andrewleith (approved):
LGTM - should be functionally equivalent, and will hopefully solve the downstream client's issue with their webserver.

mohamed-cds:
Looks like tfsec might need to be bumped. We're on `v0.39.26` while the latest is `v1.26.3`. Since we're behind this significantly this is a good opportunity to switch to checkov which is the one we've been standardizing around for terraform security scans.

cc/ @sastels since i think you looked into this in the past

jimleroyer:
@mohamed-cds Is that a low effort upgrade? just pushing the new version?

mohamed-cds:
@jimleroyer high chance it'll flag a-lot of items if you jump to a newer version in this PR. It'd recommend it as a standalone activity since you'll have to update TF resources or ignore items that we don't agree with.

I'd also switch to the official Github action instead of the one you current have. This can be done in this PR without any negative actions (famous last words).

triat/terraform-security-scan@v2.2.1 -> aquasecurity/tfsec-action@v1.0.0

mohamed-cds (approved):
Let's give this a try 🚀 ✅

PR #502, opened 2022-08-02
feat: Change document.* target group to the dd-api

sastels:
# Summary | Résumé

Files sent in the past have links that start with `https://document.notification.canada.ca/` rather than `https://api.document.notification.canada.ca/`. If we want to archive the document-download frontend then we have to make sure these old links are now handled by the document-download api.

This PR has `document.notification.canada.ca` route to document-download-api.

(in staging) change a file link from api.document.* to just document.* and verify that you still get the file.


patheard (approved):

sastels:
Followup:
This Athena query shows that we're using host document.staging.notification.cdssandbox.xyz
```
select
    from_unixtime(timestamp / 1000e0) as time,
    httprequest.clientip, header.value as host, httprequest.uri
from waf_logs
cross join UNNEST(httprequest.headers) AS t(header)
where lower(header.name)='host'
    and header.value like '%document.staging%'
    and httprequest.uri like '/d/%'
order by timestamp desc
```
And this CloudWatch query shows that we're using a document-download-api pod
```
fields @timestamp, log, kubernetes.labels.app as app
| filter kubernetes.labels.app like /document-download/
| filter @message not like /_status/
| sort @timestamp desc
| limit 20
```

PR #503, opened 2022-08-02
General waf rate limit

sastels:
# Summary | Résumé

Add a general rate limit for everything except api and dd-api.

hit admin many times until you're blocked.

### DIFF
--- a/aws/eks/waf.tf
+++ b/aws/eks/waf.tf
@@ -162,7 +162,6 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
       sampled_requests_enabled   = true
     }
 
-
     statement {
       rate_based_statement {
         limit              = 500
@@ -193,10 +192,6 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
       }
     }
   }
-
-
-
-
 
   rule {
     name     = "document_download_invalid_path"
--- a/aws/eks/waf.tf
+++ b/aws/eks/waf.tf
@@ -141,6 +141,64 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
   }
 
   rule {
+    name     = "rate_limit_all_except_api"
+    priority = 11
+
+    action {
+      block {
+        custom_response {
+          response_code = 403
+          response_header {
+            name  = "waf-block"
+            value = "RateLimitRestriction"
+          }
+        }
+      }
+    }
+
+    visibility_config {
+      cloudwatch_metrics_enabled = true
+      metric_name                = "GeneralRateLimit"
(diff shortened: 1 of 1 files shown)


aws/eks/waf.tf
```diff
@@ -204,6 +204,135 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
     }
   }
 
+  rule {
```
jimleroyer, line 207: I wonder here if that rule will apply to notification-api calling notifiacation-document-download-api?

jimleroyer (commented):
Left a question around how the rules apply.

aws/eks/waf.tf
```diff
@@ -204,6 +204,135 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+
+
+  rule {
+    name     = "rate_limit_api"
+    priority = 110
+
+    action {
+      block {
+        custom_response {
+          response_code = 403
```
mohamed-cds, line 264: @sastels any chance we can change this to a 204 so we can distinguish between geniune 4xx errors and rate limits ? Or 429 which is the one usually used for rate limits.

aws/eks/waf.tf
sastels: A good question. I think the api lambda will go through this. I guess the question is
- should we think about a way to ignore our own calls?
- should we just have a big limit instead and figure we'll be under it?
- should we have no limit on the apis?

Kinda favour the first option. Probably we can get the client ip that will come from api-lambda and allow anything from that.

It's probably a good idea to remove the api and api-lambda rules from this PR and make them (and rules for the api lambda) in a different PR.

aws/eks/waf.tf
```diff
@@ -147,7 +147,7 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
     action {
       block {
         custom_response {
-          response_code = 403
```
sastels, line 150: fixing previous PR

aws/eks/waf.tf
sastels: followup: I've removed the api rate limits from this PR and will deal with them in a separate PR.

aws/eks/waf.tf
```diff
@@ -147,7 +147,7 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+          response_code = 429
```
sastels, line 150: fixing previous PR.

jimleroyer (approved):

jimleroyer (commented):
I approved the PR but ideally, we want to set this new rule under a `count` action first so we can observe. We would switch this to `block` action after monitoring and analysis is performed.

PR #505, opened 2022-08-03
feat: geo restrict api and document-download-api

sastels:
# Summary | Résumé

https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/cds-snc/notification-planning/722

block ips from outside Canada from accessing api and document-download-api

Note: looking at our WAF logs, there were 36 requests that had no country code. Some were fuzzing attacks while the rest were just GETing `/`.  Running these through an ip lookup tool found that none were Canadian. It is reasonable to assume that legitimate traffic will have the country code properly marked.

spin up a server in AWS / us-east-1 (manually or via [terraform](https://github.com/sastels/learn-AWS-terraform/tree/main/Misc/EC2_in_us_east)) and ensure that you're blocked


jzbahrai (approved):

jimleroyer:
Regarding the usage of the API in production, normally per government policies, this should all happen within Canada. But this one will be interesting. It might have the potential to block access to vendors if any government body use these.

jimleroyer (commented):
We'll need the same rules applied to the [lambda-api waf TF file](https://github.com/cds-snc/notification-terraform/blob/d5da037344ac9b64f8948c6474fdd68db9c29137/aws/lambda-api/waf.tf#L1) as well. 

After discussion with @sastels too, we decided to change the action to `count` instead of `block` for now as we can deploy earlier in production and see how that would behave. We can reinstate the to `block` once we are ready.

PR #506, opened 2022-08-03
Block traffic matching anon ip rule set

jimleroyer:
# Summary | Résumé

Set the anon ip rule set to block by removing the count override and letting the block default take in. The following rules in the set were excluded as we observed legitimate traffic going through these:

* ~AnonymousIPList~: Not much legitimate traffic for this one, it includes bad request or at best one request coming from Tor.
* ~HostingProviderIPList~: No legitimate traffic for this one as well.

* Run the automated smoke tests.
* Manual smoke tests.


aws/eks/waf.tf
```diff
@@ -101,14 +101,22 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
     }
 
     statement {
       managed_rule_group_statement {
         name        = "AWSManagedRulesAnonymousIpList"
         vendor_name = "AWS"
       }
+
+      excluded_rule {
+        name = "AnonymousIPList"
```
jimleroyer, line 114: @patheard I didn't see much legitimate traffic coming  from that one.. except maybe for some IPs from Tor. Can we eliminate Tor related traffic? as we'd reduce some risk too?

aws/eks/waf.tf
```diff
@@ -101,14 +101,22 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+      }
+
+      excluded_rule {
+        name = "HostingProviderIPList"
```
jimleroyer, line 118: I reviewed many log pages in Athena for this one (110). I am tempted to block this. The only requests that looked valid came from scraping bots such as www.komodia.com or linkfluence.com.

aws/eks/waf.tf
```diff
@@ -101,13 +101,21 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+      none {}
     }
 
     statement {
       managed_rule_group_statement {
         name        = "AWSManagedRulesAnonymousIpList"
         vendor_name = "AWS"
+
+        excluded_rule {
+          name = "AnonymousIPList"
```
jimleroyer, line 113: I didn't see much legitimate traffic coming  from that one.. except maybe for some IPs from Tor. Can we eliminate Tor related traffic? as we'd reduce some risk too?

aws/eks/waf.tf
```diff
@@ -101,13 +101,21 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
...
+        }
+
+        excluded_rule {
+          name = "HostingProviderIPList"
```
jimleroyer, line 117: I reviewed many log pages in Athena for this one (110). I am tempted to block this. The only requests that looked valid came from scraping bots such as www.komodia.com or linkfluence.com.

aws/eks/waf.tf
mohamed-cds: i can't think of a legitimate use case where a notify user needs to anonymize their servers ip from us.

aws/eks/waf.tf
mohamed-cds: this one can cause problems since it includes a large number of cloud hosting provider IP address ranges (except AWS). So if a notify user migrated to one of these providers they wouldn't be able to interact with us.

patheard: ah interesting - I checked and Notify is the only WAF using the `AWSManagedRulesAnonymousIpList` managed ruleset, so we don't have other product data on legit requests from other cloud providers being blocked.

@mohamed-cds is totally right though - since most of the GC uses Azure, it's possible this could block legit requests in the future.

jimleroyer: I was thinking this might be smarter and not automatically include Cloud providers, but only a subset of known IPs that could be coming from these. I was naive in my assumption here? There is a lot of bad requests from this rulesets. It could be good to tweak it to apply further constraints with requests labeled with that rule, if we can't simply apply it.

patheard: Here's another way of looking at it - can you find any legit requests that are getting counted with `HostingProviderIPList`?  If no, I'd be game to try enabling it and seeing what happens.

mohamed-cds: the goal of `HostingProviderIPList` is to distinguish between user generated actions and requests originating from cloud services. Another thing we can do is block this and then open it up if we have a team that requires it. But i believe most gov services may be going through the SSC infrastructure which might never be included in this list even if they're hosted in Azure.

@patheard what do you think ?

patheard: The SSC infrastructure would for sure be excluded.  As for GC Azure hosted apps, I don't know enough about the GC infrastructure to take a guess at what IP addresses the traffic would be coming from.

The more I think about it, the more I lean towards blocking and then opening if we find a legitimate reason.

mohamed-cds: +1; let's do it @jimleroyer and revist if a trial account has issues

mohamed-cds (approved):
LGTM 🚀 ✅

aws/eks/waf.tf
jimleroyer: Just removed the excluded rule 👍

aws/eks/waf.tf
jimleroyer: Removed

patheard (approved):
🛑

PR #518, opened 2022-08-25
add ip blocklist to terraform

jzbahrai:
# Summary | Résumé

Adding the IP blocklist to our terraform repo, this way during an incident we only have to add IP's to the blocklist on console.

### DIFF
--- a/aws/eks/variables.tf
+++ b/aws/eks/variables.tf
@@ -98,5 +98,6 @@ locals {
 }
 
 variable "ip_blocklist_arn" {
-  type = string
+  description = "Block all the IPs on this list from accessing admin and api"
+  type        = string
 }
--- a/aws/eks/variables.tf
+++ b/aws/eks/variables.tf
@@ -98,5 +98,5 @@ locals {
 }
 
 variable "ip_blocklist_arn" {
-  type = list(any)
+  type = string
 }
--- a/aws/eks/variables.tf
+++ b/aws/eks/variables.tf
@@ -96,3 +96,7 @@ variable "sign_in_waf_rate_limit" {
 locals {
   eks_application_log_group = "/aws/containerinsights/${var.eks_cluster_name}/application"
 }
+
+variable "ip_blocklist" {
+  type = list(any)
+}
--- a/aws/eks/variables.tf
+++ b/aws/eks/variables.tf
@@ -97,6 +97,6 @@ locals {
   eks_application_log_group = "/aws/containerinsights/${var.eks_cluster_name}/application"
 }
 
-variable "ip_blocklist" {
+variable "ip_blocklist_arn" {
   type = list(any)
 }
--- a/aws/eks/waf.tf
+++ b/aws/eks/waf.tf
@@ -382,6 +382,27 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
     }
   }
 
+  rule {
+    name     = "ip_blocklist"
+    priority = 1
+
+    action {
+      block {}
+    }
+
+    statement {
+      ip_set_reference_statement {
+        arn = aws_wafv2_ip_set.ip_blocklist.arn
+      }
+    }
+
+    visibility_config {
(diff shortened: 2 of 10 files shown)


aws/eks/waf.tf
```diff
@@ -465,3 +486,14 @@ resource "aws_wafv2_web_acl_logging_configuration" "firehose-waf-logs" {
     }
   }
 }
+
+resource "aws_wafv2_ip_set" "ip_blocklist" {
```
jzbahrai, line 490: is there a way to put this in a common place so I am not replicating it here?
https://github.com/cds-snc/notification-terraform/pull/518/files#diff-c2256a3a91fb5b8ecd0e714c74ac644b86f2e35c8bf0cc6d3793d88c7490b071R238

sastels: You can put it in its own file in `aws/common`, add an appropriate output to `aws/common/outputs.tf`, and add it to eks like we do here
https://github.com/cds-snc/notification-terraform/blob/main/env/staging/eks/terragrunt.hcl#L62

If something fails you might have to mock it here
https://github.com/cds-snc/notification-terraform/blob/main/env/staging/eks/terragrunt.hcl#L5-L27

💯 to making one file and sharing in the two wafs

jzbahrai: thank you

env/staging/eks/terragrunt.hcl
```diff
@@ -7,7 +7,8 @@ dependency "common" {
 
   # Configure mock outputs for the `validate` command that are returned when there are no outputs available (e.g the
   # module hasn't been applied yet.
-  mock_outputs_allowed_terraform_commands = ["validate"]
+  mock_outputs_allowed_terraform_commands = ["init", "fmt", "validate", "plan", "show"]
+  mock_outputs_merge_with_state           = true
```
sastels, line 11: don't need these in the lambda terragrunt file?

jzbahrai: their there, i copy pasted from that

sastels:
hrm, these are empty - strange?
```
              + ip_set_reference_statement {
                }
```

jzbahrai:
This is was @patheard was explaining to me. Because I added the resource for ip_blocklist in the same PR as adding to the staging and prod env, the result gets mocked out. If I had added the rule in another PR it wouldn't be empty. @sastels

aws/eks/variables.tf
```diff
@@ -96,3 +96,7 @@ variable "sign_in_waf_rate_limit" {
 locals {
   eks_application_log_group = "/aws/containerinsights/${var.eks_cluster_name}/application"
 }
+
+variable "ip_blocklist_arn" {
+  type = string
```
patheard, line 101: Could you add a variable `description` explaining what this is used for.

aws/lambda-api/variables.tf
```diff
@@ -104,3 +104,7 @@ variable "new_relic_account_id" {
 variable "ff_cloudwatch_metrics_enabled" {
   type = bool
 }
+
+variable "ip_blocklist_arn" {
+  type = string
```
patheard, line 109: nit: same comment about adding a `description`

env/production/eks/terragrunt.hcl
```diff
@@ -45,4 +45,5 @@ inputs = {
   eks_node_ami_version                   = "1.22.9-20220725"
   fall_back_non_api_waf_rate_limit       = 500
   sign_in_waf_rate_limit                 = 100
+  ip_blocklist_arn                       = dependency.common.outputs.ip_blocklist_arn
```
patheard, line 48: This will likely fail when you go to do the `Production` PR since there are no mocked outputs for the `common` dependency, so the `terraform plan` will error out.

Simple fix would be to add the same mocked outputs you have for `env/staging/eks/terragrunt.hcl`

patheard (approved):

PR #516, opened 2022-08-23
Dont geoblock us

sastels:
# Summary | Résumé

Seeing valid traffic from US to api, so we can't geoblock the US.
[CanadaOnlyGeoRestriction Analysis](https://docs.google.com/spreadsheets/d/1aIB8QlbyRR5Ex96I0qgEdnXqbYQ9hlxYoXQKaTN-cVk/edit#gid=332050588)

# Test instructions | Instructions pour tester la modification
Run the analysis again (though in the above analysis only US traffic was legit)

PR #519, opened 2022-08-25
fix: Don't geoblock US for api lambda waf too

sastels:
# Summary | Résumé

Forgot to change the api lambda's WAF in #516

- After merging to staging
- Do a GET to https://api.staging.notification.cdssandbox.xyz  from servers in CA, US, and EU.
- Then in Athena run
```
with data as (
    select 
        from_unixtime(timestamp / 1000e0) as time, action, nonterminatingmatchingrule,
        httprequest.clientip, httprequest.country, httprequest.httpmethod, hostheader.value as host, httprequest.uri
    from waf_logs
    cross join UNNEST(httprequest.headers) AS t(hostheader)
    cross join UNNEST(nonterminatingmatchingrules) AS t(nonterminatingmatchingrule)
    where hostheader.name = 'Host'
)
select * from data
where clientip = '...'
order by time desc
```
should have the `CanadaUSOnlyGeoRestriction` counted for EU but not CA or US.

### DIFF
--- a/aws/lambda-api/waf.tf
+++ b/aws/lambda-api/waf.tf
@@ -151,7 +151,7 @@ resource "aws_wafv2_web_acl" "api_lambda" {
       not_statement {
         statement {
           geo_match_statement {
-            country_codes = ["CA"]
+            country_codes = ["CA", "US"]
           }
         }
       }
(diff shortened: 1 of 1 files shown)


patheard (approved):

sastels:
Followup: tested in staging, now counting for EU but not CA / US

PR #536, opened 2022-09-20
Removed fargate profile

jimleroyer:
# Summary | Résumé

Fargate profile support has been dropped for a few AZ in ca-central-1. As we are not currently using it, dropping its creation will be simpler. We can revive it later if we decide to move our kubernetes out of EC2 and into fargate.

Terraform plan and execution should now work (currently failing).


jzbahrai (approved):

sastels:
according to the terraform plan action there aren't any changes - confused...

PR #540, opened 2022-10-05
Added base path mapping for api lambda from alt domain name

jimleroyer:
# Summary | Résumé

* Added base path mapping for api lambda from alt domain name

Try the domain api.staging.notification.alpha.cdssandbox.xyz !


patheard (approved):
🐉 🐉 🐉

PR #541, opened 2022-10-05
Enabling bot control for both eks and lambda api

jimleroyer:
# Summary | Résumé

Not to be merged yet. I will run a traffic analysis to check for any potential traffic we might block if we enable this ruleset.

* Enabling bot control for both eks and lambda api

* Monitor the traffic for any legitimate traffic that could get blocked.


jimleroyer:
We are not going forward with implementing bot control, as these are a bit more complicated to distinguish legit (web archival services, etc) vs attacks (there are no clear signs from the request). We will wait for hardening these, when patterns are recognizable and which can be derived from these rules.

PR #546, opened 2022-10-07
Initial blazer commit

jzbahrai:
# Summary | Résumé

Setting up blazer using terraform. 
huge thanks to @sastels and @patheard

### DIFF
--- a/.github/workflows/terragrunt_plan_staging.yml
+++ b/.github/workflows/terragrunt_plan_staging.yml
@@ -93,6 +93,9 @@ jobs:
             heartbeat:
               - 'aws/heartbeat/**'
               - 'env/staging/heartbeat/**'
+            database-tools:
+              - 'aws/database-tools/**'
+              - 'env/staging/database-tools/**'
       
       - name: Terragrunt plan common
         if: ${{ steps.filter.outputs.common == 'true' }}
--- a/.github/workflows/terragrunt_plan_staging.yml
+++ b/.github/workflows/terragrunt_plan_staging.yml
@@ -196,3 +196,13 @@ jobs:
           comment-title: "Staging: heartbeat"
           github-token: "${{ secrets.GITHUB_TOKEN }}"
           terragrunt: "true"
+
+      - name: Terragrunt plan database-tools
+        if: ${{ steps.filter.outputs.database-tools == 'true' || steps.filter.outputs.common == 'true' }}
+        uses: cds-snc/terraform-plan@v1
+        with:
+          directory: "env/staging/database-tools"
+          comment-delete: "true"
+          comment-title: "Staging: database-tools"
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+          terragrunt: "true"
--- a/aws/database-tools/blazer.tf
+++ b/aws/database-tools/blazer.tf
@@ -26,6 +26,10 @@ resource "aws_ecs_task_definition" "blazer" {
   network_mode             = "awsvpc"
   requires_compatibilities = ["FARGATE"]
 
+  vars = {
+    DATABASE_URL = aws_ssm_parameter.sqlalchemy_database_reader_uri.arn
+  }
+
   cpu    = 256
   memory = 512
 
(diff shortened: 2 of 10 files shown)


aws/database-tools/blazer.tf
```diff
@@ -0,0 +1,63 @@
...
+  task_definition        = aws_ecs_task_definition.blazer.arn
+  desired_count          = 1
+  launch_type            = "FARGATE"
+  enable_execute_command = true
+
+  network_configuration {
+    security_groups = [aws_security_group.blazer.id]
+    subnets         = var.vpc_private_subnets
+  }
+}
```
jimleroyer, line 22: So Blaser will run as an elastic container service, deployed via Fargate and running within our VPC?

aws/database-tools/variables.tf
```diff
@@ -0,0 +1,21 @@
+variable "vpc_private_subnets" {
+  type = list(any)
+}
+
+variable "vpc_id" {
+  type = string
+}
+
+variable "sqlalchemy_database_reader_uri" {
```
jimleroyer, line 9: I'd remove reference to SQLAlchemy in that database URI. I'd assume it's unimportant to know here, unless it is for reasons I missed?

aws/database-tools/blazer.tf
jzbahrai: yes

aws/database-tools/variables.tf
patheard: If you create a secret or `SecureString` ParameterStore, you can then feed that into your Blazer container definition as a secret, which will let you see your container definition in the `terraform plans` rather than having the whole thing marked as `(sensitive)`:
https://github.com/cds-snc/security-tools/blob/main/terragrunt/aws/software_asset_inventory/container-definitions/dependencytrack_api.json.tmpl#L71-L76

The value being passed in here is a ParameterStore ARN:
https://github.com/cds-snc/security-tools/blob/main/terragrunt/aws/software_asset_inventory/api_server.tf#L35

.github/workflows/terragrunt_plan_staging.yml
```diff
@@ -93,6 +93,9 @@ jobs:
             heartbeat:
               - 'aws/heartbeat/**'
               - 'env/staging/heartbeat/**'
+            database-tools:
```
patheard, line 96: I think you'll want to also update the `Merge to main (staging)` workflow so the `terraform apply` runs:
https://github.com/cds-snc/notification-terraform/blob/main/.github/workflows/merge_to_main_staging.yml

aws/database-tools/securitygroups.tf
```diff
@@ -0,0 +1,22 @@
+resource "aws_security_group" "blazer" {
```
patheard, line 1: You'll also need an egress that lets the ECS task talk to the database like so:
https://github.com/cds-snc/security-tools/blob/main/terragrunt/aws/software_asset_inventory/security_groups.tf#L22-L28

aws/database-tools/task_role.tf
```diff
@@ -0,0 +1,22 @@
+resource "aws_iam_role" "blazer_ecs_task" {
+  name               = "${aws_ecs_cluster.blazer.name}-ecs-task"
+  assume_role_policy = data.aws_iam_policy_document.ecs_task_assume.json
+  tags = {
+    (var.billing_tag_key) = var.billing_tag_value
+  }
+}
+
+resource "aws_iam_role_policy_attachment" "blazer_ecs_task_policy_attach" {
```
patheard, line 9: An IAM policy you'll want on this is to allow for the port forwarding:
https://github.com/mohamed-cds/terraform_test_infrastructure/blob/main/ssm_private_subnet/terraform/iam.tf#L66-L77

sastels (approved):
Let's try this out!

aws/database-tools/variables.tf
jimleroyer: I see, we could rename this once we move to ParameterStore instead.

PR #567, opened 2022-10-21
wrong engine, needed to be fixed

jzbahrai:
# Summary | Résumé

The name of the engine needed to be "postgres" and not "PostgreSQL"

this is how I found the right engine: https://github.com/terraform-aws-modules/terraform-aws-rds/blob/v5.1.0/examples/complete-postgres/main.tf#L72


andrewleith (approved):
LGTM!

PR #570, opened 2022-10-24
Increase SMS limit on Staging from $30 -> $50

smcmurtry:
# Summary | Résumé

We have already been granted the increase from AWS. Increase SMS limit on Staging from $30 -> $50.


sastels (approved):
🎉

PR #571, opened 2022-10-24
fix permission for blazer execution role

jzbahrai:
# Summary | Résumé

trying to fix this:image

this Pr failed: https://github.com/cds-snc/notification-terraform/pull/568
because we cant have both resources and principals attached the the same policy

### DIFF
--- a/aws/database-tools/container_execution_role.tf
+++ b/aws/database-tools/container_execution_role.tf
@@ -35,3 +35,43 @@ data "aws_iam_policy_document" "blazer_execution_role" {
     }
   }
 }
+
+resource "aws_iam_policy" "blazer_exection_role_parameter_policy" {
+  name   = "blazer_exection_role_parameter_policy"
+  path   = "/"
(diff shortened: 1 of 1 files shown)


aws/database-tools/container_execution_role.tf
```diff
@@ -35,3 +35,43 @@ data "aws_iam_policy_document" "blazer_execution_role" {
...
+
+    effect = "Allow"
+
+    actions = [
+      "ssmmessages:CreateControlChannel",
+      "ssmmessages:CreateDataChannel",
+      "ssmmessages:OpenControlChannel",
+      "ssmmessages:OpenDataChannel"
+    ]
+    resources = ["*"]
```
sastels, line 61: Can we restrict this at all?

sastels (approved):
Could restrict that statement if possible, otherwise :shipit:

aws/database-tools/container_execution_role.tf
jzbahrai: o i merged, but yes! will restrict further.. I'm actually unsure if we need this permission , but i can change it after

PR #572, opened 2022-10-24
Rename parameter storevariables and update prod plans

jzbahrai:
# Summary | Résumé

Originally:
BLAZER_DATABASE_URL -> Database we added for database tools
DATABASE_URL -> notification db

This was incorrect. We had to switch it. 
I clickopsed and tried it out and saw if this was the case (it was).

Now I renamed variables in parameter store

BLAZER_DATABASE -> DATABASE_TOOLS_URL (and then we call it DATABASE_URL) in the DB Cast
sqlalchemy_database_reader_uri -> NOTIFICATION_DB_READ_URL

### DIFF
--- a/aws/database-tools/blazer.tf
+++ b/aws/database-tools/blazer.tf
@@ -54,10 +54,10 @@ resource "aws_ecs_task_definition" "blazer" {
         }
       ],
       "secrets" : [{
-        "name" : "DATABASE_URL",
+        "name" : "BLAZER_DATABASE_URL",
         "valueFrom" : "${aws_ssm_parameter.sqlalchemy_database_reader_uri.arn}"
         }, {
-        "name" : "BLAZER_DATABASE_URL",
+        "name" : "DATABASE_URL",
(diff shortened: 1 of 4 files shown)


aws/database-tools/blazer.tf
```diff
@@ -54,10 +54,10 @@ resource "aws_ecs_task_definition" "blazer" {
         }
       ],
       "secrets" : [{
-        "name" : "DATABASE_URL",
+        "name" : "BLAZER_DATABASE_URL",
         "valueFrom" : "${aws_ssm_parameter.sqlalchemy_database_reader_uri.arn}"
         }, {
-        "name" : "BLAZER_DATABASE_URL",
+        "name" : "DATABASE_URL",
```
sastels, line 60: Why did these get swapped?

jzbahrai:
@sastels yes I think so regarding security. its all tied to security groups that are getting connections from the eks security group and vice versa.

sastels (approved):
LGTM! Should we make any changes to our ATO docs? (ie pointing out a new server with access to our data)

PR #573, opened 2022-10-26
feat: database tools ECR

patheard:
# Summary
Add an ECR to hold the database tools images and remove the ingress
on port 8080 as this is handled automatically by the SSM Session
Manager.

The PR also swaps the ECS task env vars, which is a change that has
already been made in the AWS console.

# Related
- #572
- cds-snc/platform-core-services#137

### DIFF
--- a/aws/database-tools/ecr.tf
+++ b/aws/database-tools/ecr.tf
@@ -1,0 +1,12 @@
+resource "aws_ecr_repository" "blazer" {
+  name                 = "database-tools/blazer"
+  image_tag_mutability = "IMMUTABLE"
+
+  image_scanning_configuration {
+    scan_on_push = true
+  }
+
+  tags = {
+    CostCenter = "notification-canada-ca-${var.env}"
+  }
+}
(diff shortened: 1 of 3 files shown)


aws/database-tools/ecr.tf
```diff
@@ -0,0 +1,12 @@
+resource "aws_ecr_repository" "blazer" {
+  name                 = "database-tools/blazer"
```
jzbahrai, line 2: lgtm! https://ca-central-1.console.aws.amazon.com/ecr/repositories/private/239043911459/database-tools/blazer?region=ca-central-1

do we change this to `database-tools/blazer:v2.6.5`

jzbahrai (approved):

aws/database-tools/ecr.tf
patheard: we'll add a tagged `v2.6.5` image to the ECR so the ECS task will end up referencing an image that looks like so:
```
arn:aws:ecr:ca-central-1:$ACCOUNT_ID:database-tools/blazer:v2.6.5
```

PR #577, opened 2022-11-10
Implement Private links

jzbahrai:
# Summary | Résumé

https://docs.aws.amazon.com/vpc/latest/privatelink/aws-services-privatelink-support.html

These are the services I added:
https://github.com/cds-snc/notification-terraform/pull/577/files#diff-6e098f15ef7306c0a06380c16a90de9014fbe5744e94f0f8835a50a98e8e3e67R2-R7

Let me know if you think I missed something:
1.AWS  ECR
2. AWS Cloudwatch Logs
3. AWS Cloudwatch Events
4. Amazon Cloudwatch
5. Amazon RDS
6. AWS lambda
7. Amazon ElasticCache
8. AWS Secrets Manager
9. Amazon SES
10. Amazon SNS
11. Amazon SQS

### DIFF
--- a/aws/common/outputs.tf
+++ b/aws/common/outputs.tf
@@ -63,7 +63,9 @@ output "private-links-vpc-endpoints-securitygroup" {
   description = "private links vpc endpoint security group id"
 }
 
-output "private-links-gateway" {
-  value       = aws_vpc_endpoint.gateway
-  description = "private links gateway vc endpoint"
+output "private-links-gateway-prefix-list-ids" {
+  value = [
+    for gateway in aws_vpc_endpoint.gateway : gateway.prefix_list_id
+  ]
+  description = "The prefix list IDs for the gateway private links"
 }
--- a/aws/common/outputs.tf
+++ b/aws/common/outputs.tf
@@ -57,3 +57,13 @@ output "firehose_waf_logs_iam_role_arn" {
 output "ip_blocklist_arn" {
   value = aws_wafv2_ip_set.ip_blocklist.arn
 }
+
+output "private-links-vpc-endpoints-securitygroup" {
+  value       = aws_security_group.vpc_endpoints.id
+  description = "private links vpc endpoint security group id"
+}
+
+output "private-links-gateway" {
+  value       = aws_vpc_endpoint.gateway
+  description = "private links gateway vc endpoint"
+}
--- a/aws/common/private_links.tf
+++ b/aws/common/private_links.tf
@@ -2,9 +2,8 @@
 # TODO: add these later - "lambda", "elasticache", "elasticache-fips", "email-smtp", "sns", "sqs",
 locals {
   endpoints_interface = toset([
-    "ecr.dkr", "ecr.api", "logs", "events",
-    "evidently", "evidently-dataplane", "monitoring", "rum", "rum-dataplane", "synthetics",
-    "rds", "secretsmanager"
+    "ecr.dkr", "ecr.api", "logs", "ssm",
+    "ssmmessages", "ec2messages", "monitoring", "rds"
   ])
   endpoints_gateway = toset(["s3"])
 }
@@ -19,7 +18,7 @@ resource "aws_vpc_endpoint" "interface" {
   security_group_ids = [
     aws_security_group.vpc_endpoints.id,
   ]
-  subnet_ids = aws_subnet.notification-canada-ca-private.*.id
+  subnet_ids = aws_subnet.notification-canada-ca-private[*].id
 }
 
 resource "aws_vpc_endpoint" "gateway" {
@@ -28,11 +27,7 @@ resource "aws_vpc_endpoint" "gateway" {
   vpc_id            = aws_vpc.notification-canada-ca.id
   vpc_endpoint_type = "Gateway"
   service_name      = "com.amazonaws.${var.region}.${each.value}"
-  route_table_ids = [
-    aws_route_table.notification-canada-ca-private_subnet[0].id,
-    aws_route_table.notification-canada-ca-private_subnet[1].id,
-    aws_route_table.notification-canada-ca-private_subnet[2].id
-  ]
+  route_table_ids   = aws_route_table.notification-canada-ca-private_subnet[*].id
 }
 
 resource "aws_security_group" "vpc_endpoints" {
--- a/aws/common/private_links.tf
+++ b/aws/common/private_links.tf
@@ -1,0 +1,37 @@
+locals {
+  endpoints_interface = toset([
+    "ecr.dkr", "ecr.api", "logs", "events",
+    "evidently", "evidently-dataplane", "monitoring", "rum", "rum-dataplane", "synthetics",
+    "rds", "lambda", "elasticache", "elasticache-fips",
+    "secretsmanager", "email-smtp", "sns", "sqs",
+  ])
+  endpoints_gateway = toset(["s3"])
+}
+
+resource "aws_vpc_endpoint" "interface" {
+  for_each = local.endpoints_interface
+
+  vpc_id              = aws_vpc.notification-canada-ca.id
+  vpc_endpoint_type   = "Interface"
+  service_name        = "com.amazonaws.${var.region}.${each.value}"
+  private_dns_enabled = true
+  security_group_ids = [
+    aws_security_group.vpc_endpoints.id,
+  ]
+  subnet_ids = aws_subnet.notification-canada-ca-private.*.id
+}
+
+resource "aws_vpc_endpoint" "gateway" {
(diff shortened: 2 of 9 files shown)


aws/common/private_links.tf
```diff
@@ -0,0 +1,37 @@
...
+  for_each = local.endpoints_interface
+
+  vpc_id              = aws_vpc.notification-canada-ca.id
+  vpc_endpoint_type   = "Interface"
+  service_name        = "com.amazonaws.${var.region}.${each.value}"
+  private_dns_enabled = true
+  security_group_ids = [
+    aws_security_group.vpc_endpoints.id,
+  ]
+  subnet_ids = aws_subnet.notification-canada-ca-private.*.id
```
jzbahrai, line 21: should this be the private subnet ids?

aws/common/private_links.tf
```diff
@@ -0,0 +1,37 @@
...
+}
+
+resource "aws_vpc_endpoint" "gateway" {
+  for_each = local.endpoints_gateway
+
+  vpc_id            = aws_vpc.notification-canada-ca.id
+  vpc_endpoint_type = "Gateway"
+  service_name      = "com.amazonaws.${var.region}.${each.value}"
+  route_table_ids   = [aws_route_table.notification-canada-ca-public_subnet.id]
```
jzbahrai, line 30: is this the correct route table?

aws/eks/securitygroups.tf
```diff
@@ -106,3 +106,187 @@ resource "aws_security_group" "notification-canada-ca-worker" {
...
   }
 }
+
+
+###
+# Connect all security groups to go through the private link
+###
+
+# the load balancer
+resource "aws_security_group_rule" "loadbalancer-egress-private-endpoints" {
```
jzbahrai, line 116: for each existing security group there are 3 rules:
1. egress to vpc endpoints from *service
2. vpc endpoint ingress from *service
3. service* egres to gateway (for s3)

aws/eks/securitygroups.tf
```diff
@@ -106,3 +106,187 @@ resource "aws_security_group" "notification-canada-ca-worker" {
...
+  type                     = "ingress"
+  from_port                = 443
+  to_port                  = 443
+  protocol                 = "tcp"
+  source_security_group_id = aws_security_group.notification-canada-ca-alb.id
+  security_group_id        = var.private-links-vpc-endpoints-securitygroup
+}
+
+resource "aws_security_group_rule" "notify-lb-egress-endpoints-gateway" {
+  for_each = toset(var.private-links-gateway)
```
jzbahrai, line 137: I left this as a list and forced it to be a set (tf plan required it) though the only thing under gateway is S3? so maybe this needs to change

aws/common/private_links.tf
patheard: yup, exactly

aws/common/private_links.tf
patheard: no, it should come from the VPC like so:
```hcl
aws_vpc.notification-canada-ca.main_route_table_id
```

aws/common/private_links.tf
```diff
@@ -0,0 +1,37 @@
+locals {
+  endpoints_interface = toset([
+    "ecr.dkr", "ecr.api", "logs", "events",
+    "evidently", "evidently-dataplane", "monitoring", "rum", "rum-dataplane", "synthetics",
+    "rds", "lambda", "elasticache", "elasticache-fips",
+    "secretsmanager", "email-smtp", "sns", "sqs",
```
patheard, line 6: It might make your testing easier if you only add the VPC endpoints needed for Blazer and then add in the others in a future PR.  I'm not entirely sure which endpoints you'll need for all the Notify services.

aws/eks/securitygroups.tf
patheard: I think this'll be ok, but if it doesn't work on the apply, we can just hard code it to only be the `s3` endpoint.  The only other type of gateway endpoint is `dynamodb` which I don't think Notify is planning to use any time soon.

aws/common/outputs.tf
```diff
@@ -57,3 +57,13 @@ output "firehose_waf_logs_iam_role_arn" {
...
   value = aws_wafv2_ip_set.ip_blocklist.arn
 }
+
+output "private-links-vpc-endpoints-securitygroup" {
+  value       = aws_security_group.vpc_endpoints.id
+  description = "private links vpc endpoint security group id"
+}
+
+output "private-links-gateway" {
+  value       = aws_vpc_endpoint.gateway
```
patheard, line 67: Unfortunately you won't be able to set a resource as an output, but you can return a list of the prefix IDs like so:
```hcl
output "private-links-gateway-prefix-list-ids" {
  value = [
    for gateway in aws_vpc_endpoint.gateway : gateway.prefix_list_id
  ]
  description = "The prefix list IDs for the gateway private links"
}
```

aws/eks/securitygroups.tf
patheard: I double-checked how you were setting the gateway resources as an output and realized it wasn't going to work.  I've updated your output so now it's just a list of prefix list IDs and can be used like so:
```hcl
resource "aws_security_group_rule" "notify-lb-egress-endpoints-gateway" {
  description       = "Security group rule for notification loadbalancer to S3 gateway"
  type              = "egress"
  from_port         = 443
  to_port           = 443
  protocol          = "tcp"
  security_group_id = aws_security_group.notification-canada-ca-alb.id
  prefix_list_ids   = var.private-links-gateway-prefix-list-ids
}
```

jzbahrai: thanks! I changed this in all the places there was a rule for the gateway

aws/common/outputs.tf
jzbahrai: done

aws/common/private_links.tf
jzbahrai: changed this to [aws_route_table.notification-canada-ca-private_subnet[0].id, aws_route_table.notification-canada-ca-private_subnet[1].id, aws_route_table.notification-canada-ca-private_subnet[2].id] 
as it had a count of 3

aws/common/private_links.tf
jzbahrai: removed everything other than what I think is needed for blazer.

aws/common/private_links.tf
```diff
@@ -0,0 +1,42 @@
+
+# TODO: add these later - "lambda", "elasticache", "elasticache-fips", "email-smtp", "sns", "sqs",
+locals {
+  endpoints_interface = toset([
+    "ecr.dkr", "ecr.api", "logs", "events",
+    "evidently", "evidently-dataplane", "monitoring", "rum", "rum-dataplane", "synthetics",
+    "rds", "secretsmanager"
```
patheard, line 7: Could you tighten this up even more to just use these interface endpoints:
```
"ecr.dkr", "ecr.api", "logs", "ssm", "ssmmessages", "ec2messages"
```
This should be the lightest touch change to support Blazer and the port forwarding session.

aws/common/private_links.tf
```diff
@@ -0,0 +1,42 @@
...
+  for_each = local.endpoints_gateway
+
+  vpc_id            = aws_vpc.notification-canada-ca.id
+  vpc_endpoint_type = "Gateway"
+  service_name      = "com.amazonaws.${var.region}.${each.value}"
+  route_table_ids = [
+    aws_route_table.notification-canada-ca-private_subnet[0].id,
+    aws_route_table.notification-canada-ca-private_subnet[1].id,
+    aws_route_table.notification-canada-ca-private_subnet[2].id
+  ]
```
patheard, line 35: You can tighten this up with a splat expression:
```hcl
route_table_ids = aws_route_table.notification-canada-ca-private_subnet[*].id
```

aws/common/private_links.tf
```diff
@@ -0,0 +1,42 @@
...
+  for_each = local.endpoints_interface
+
+  vpc_id              = aws_vpc.notification-canada-ca.id
+  vpc_endpoint_type   = "Interface"
+  service_name        = "com.amazonaws.${var.region}.${each.value}"
+  private_dns_enabled = true
+  security_group_ids = [
+    aws_security_group.vpc_endpoints.id,
+  ]
+  subnet_ids = aws_subnet.notification-canada-ca-private.*.id
```
patheard, line 22: If you're curious, this is a legacy splat expression, which I just learned was legacy.  The updated version I just proposed below would work here as well:
```hcl
subnet_ids = aws_subnet.notification-canada-ca-private[*].id
```

patheard (approved):
🧪

PR #578, opened 2022-11-14
test: update Blazer to use Google Authentication

patheard:
# Summary
Update the Blazer ECS task to use a Docker image that is setup to automatically authenticate users with their Google account when they connect.

This will give us an audit table that shows the queries run by users.

# :warning: Note
Once this is applied, we'll need to connect to the Blazer container and create the new `Users` database table with:
```sh
rails db:migrate
```

# Related
- cds-snc/platform-core-services#137


jzbahrai (approved):
🚀

PR #579, opened 2022-11-16
Testing blazer without access to the internet

jzbahrai:
# Summary | Résumé

We want to see if we can access blazer without having to exit the vpc. This PR will help us with that.


jimleroyer (approved):
I have no clue but I understand it's for testing purpose and it does not affect production.

PR #582, opened 2022-11-17
Lambda for google cidr

jzbahrai:
# Summary | Résumé

Lambda for google cidr

### DIFF
--- a/aws/lambda-google-cidr/lambda.tf
+++ b/aws/lambda-google-cidr/lambda.tf
@@ -1,0 +1,40 @@
+module "lambda-google-cidr" {
+  source                 = "github.com/cds-snc/terraform-modules?ref=v0.0.45//lambda"
+  name                   = "google-cidr"
+  billing_tag_value      = var.billing_tag_value
+  ecr_arn                = aws_ecr_repository.google-cidr.arn
+  enable_lambda_insights = true
+  image_uri              = "${aws_ecr_repository.google-cidr.repository_url}:latest"
+  timeout                = 60
+  memory                 = 1024
+
+  environment_variables = {
+    PREFIX_LIST_ID = var.google_cidr_prefix_list_id
+  }
+}
+
+resource "aws_lambda_function_event_invoke_config" "google_cidr_invoke_config" {
+  function_name                = module.lambda-google-cidr.function_name
+  maximum_event_age_in_seconds = 60
+  maximum_retry_attempts       = 0
+}
+
+resource "aws_cloudwatch_event_target" "google_cidr" {
+  arn  = module.lambda-google-cidr.function_arn
+  rule = aws_cloudwatch_event_rule.google_cidr.id
+}
+
+resource "aws_cloudwatch_event_rule" "google_cidr" {
+  name                = "google_cidr_testing"
+  description         = "google_cidr_testing event rule"
+  schedule_expression = var.google_cidr_schedule_expression
+  depends_on          = [module.lambda-google-cidr]
+}
+
+resource "aws_lambda_permission" "allow_cloudwatch" {
+  statement_id  = "AllowExecutionFromCloudWatch"
+  action        = "lambda:InvokeFunction"
+  function_name = module.lambda-google-cidr.function_name
+  principal     = "events.amazonaws.com"
+  source_arn    = aws_cloudwatch_event_rule.google_cidr.arn
+}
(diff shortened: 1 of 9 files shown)


env/staging/lambda-google-cidr/terragrunt.hcl
```diff
@@ -0,0 +1,33 @@
...
+  }
+}
+
+include {
+  path = find_in_parent_folders()
+}
+
+inputs = {
+  billing_tag_value                  = "notification-canada-ca-staging"
+  google_cidr_schedule_expression    = "rate(5 minutes)"
```
jzbahrai, line 27: ill change this after, this is so we can test atm

aws/lambda-google-cidr/lambda.tf
```diff
@@ -0,0 +1,40 @@
+module "lambda-google-cidr" {
+  source                 = "github.com/cds-snc/terraform-modules?ref=v0.0.45//lambda"
```
patheard, line 2: Can you change `v0.0.45` to the latest tag release from the terraform-modules repo.

patheard (changes requested):
Thanks for taking care of this PR!

A few tweaks to add:
1. Can you create the [managed prefix list](https://github.com/cds-snc/notification-terraform/blob/feat/google-service-ip-prefix-list/aws/eks/securitygroups.tf#L70-L90) for the Google CIDRs and associate it with the Blazer security group.
2. We'll also need a [custom IAM policy](https://github.com/cds-snc/notification-terraform/blob/feat/google-service-ip-prefix-list/aws/eks/securitygroups.tf#L70-L90) added to the lambda function that allows it to manage the prefix list.

aws/lambda-google-cidr/lambda.tf
```diff
@@ -0,0 +1,116 @@
...
+    effect = "Allow"
+    actions = [
+      "sts:AssumeRole",
+    ]
+    principals {
+      type        = "Service"
+      identifiers = ["lambda.amazonaws.com"]
+    }
+  }
+}
```
patheard, line 81: This is exactly right for how normally create an IAM role + policies but for our `lambda` module it's done a little differently to you can delete all these resources.

aws/lambda-google-cidr/lambda.tf
```diff
@@ -0,0 +1,116 @@
...
+
+data "aws_iam_policy_document" "google_cidrs" {
```
patheard, line 83: You just need to specify this element as a custom policy to attach to the lambda func module:
```hcl
policies = [
  data. aws_iam_policy_document.google_cidrs.json
]
```
Here's an example of how we're doing it in another project:
https://github.com/cds-snc/scan-files/blob/main/terragrunt/aws/api/lambda.tf#L29-L33

aws/lambda-google-cidr/lambda.tf
```diff
@@ -0,0 +1,116 @@
...
+    effect = "Allow"
+    actions = [
+      "logs:CreateLogStream",
+      "logs:PutLogEvents"
+    ]
+    resources = [
+      aws_cloudwatch_log_group.google_cidrs.arn,
+      "${aws_cloudwatch_log_group.google_cidrs.arn}:log-stream:*"
+    ]
+  }
```
patheard, line 94: You can remove this since it gets created for us by the `lambda` module.

aws/lambda-google-cidr/lambda.tf
```diff
@@ -0,0 +1,116 @@
...
+  source_arn    = aws_cloudwatch_event_rule.google_cidr.arn
+}
+
+resource "aws_cloudwatch_log_group" "google_cidrs" {
+  name              = module.lambda-google-cidr.function_name
+  retention_in_days = 7
+  tags = {
+    CostCentre = "notification-canada-ca-${var.env}"
+  }
+}
```
patheard, line 50: You can remove this since the `lambda` module creates it as well.

patheard:
Something else I just noticed - could you change the Blazer ECR to `MUTABLE` as well:
https://github.com/cds-snc/notification-terraform/blob/main/aws/database-tools/ecr.tf#L3

aws/lambda-google-cidr/lambda.tf
jzbahrai: done

aws/lambda-google-cidr/lambda.tf
jzbahrai: i removed these 4 resources

aws/lambda-google-cidr/lambda.tf
jzbahrai: done

aws/lambda-google-cidr/lambda.tf
jzbahrai: i removed this one statement but kept the rest

patheard (approved):

PR #595, opened 2022-12-08
Setup ses_receiving_emails lambda

jzbahrai:
# Summary | Résumé

Setup ses_receiving_emails lambda


.github/workflows/merge_to_main_staging.yml
```diff
@@ -24,6 +24,10 @@ env:
   TF_VAR_heartbeat_api_key: ${{ secrets.STAGING_HEARTBEAT_API_KEY }}
   TF_VAR_heartbeat_base_url: ${{ secrets.STAGING_HEARTBEAT_BASE_URL }}
   TF_VAR_heartbeat_template_id: ${{ secrets.STAGING_HEARTBEAT_TEMPLATE_ID }}
+  TF_VAR_notify_sending_domain: ${{ secrets.STAGING_NOTIFY_SENDING_DOMAIN}}
+  TF_VAR_sqs_region: ${{ secrets.STAGING_SQS_REGION }}
+  TF_VAR_celery_queue_prefix: ${{ secrets.STAGING_CELERY_QUEUE_PREFIX }}
+  TF_VAR_gc_notify_service_email: ${{ secrets.STAGING_GC_NOTIFY_SERVICE_EMAIL }}
```
patheard, line 30: Are all these secrets or could some of them be hardcoded as Terraform variables in the `terragrunt.hcl` files?

aws/ses_receiving_emails/cloudwatch_logs.tf
```diff
@@ -0,0 +1,25 @@
...
+
+resource "aws_cloudwatch_log_group" "ses_receiving_emails_log_group" {
+  name              = "ses_receiving_emails_log_group"
+  retention_in_days = 90
+  tags = {
+    CostCenter  = "notification-canada-ca-${var.env}"
+    Environment = var.env
+    Application = "lambda"
+  }
+}
```
patheard, line 13: You should be able to get rid of this log group since the Terraform lambda module will create one for the function automagically.

aws/ses_receiving_emails/lambda.tf
```diff
@@ -0,0 +1,43 @@
...
+  image_uri              = "${aws_ecr_repository.ses_receiving_emails.repository_url}:latest"
+  timeout                = 60
+  memory                 = 1024
+
+  environment_variables = {
+    NOTIFY_SENDING_DOMAIN   = var.notify_sending_domain
+    SQS_REGION              = var.sqs_region
+    CELERY_QUEUE_PREFIX     = var.celery_queue_prefix
+    GC_NOTIFY_SERVICE_EMAIL = var.gc_notify_service_email
+  }
```
patheard, line 16: Another note about secret values - if any of these values MUST NOT BE REVEALED, it's better not to use the `environment_variables` block since AWS cli commands will leak these values.

The pattern we've used on other functions is:
1. Put the secret values in SecretsManager or ParameterStore
2. During lambda startup, load the secrets into the function's execution environment.

I can show you an example if you decide to change how any of these are loaded.

aws/ses_receiving_emails/variables.tf
```diff
@@ -0,0 +1,38 @@
...
+}
+
+variable "celery_queue_prefix" {
+  type        = string
+  description = "Celery queue prefix"
+}
+
+variable "gc_notify_service_email" {
+  type        = string
+  description = "Sevice email for GC Notify."
```
jimleroyer, line 24:
```suggestion
  description = "Service email for GC Notify."
```

jimleroyer (commented):
I forgot to add, but we talked about potentially moving the ses receive email lambda into the Canadian AWS region. Unfortunately, we'll have to keep it into the American region for now. That is because reception of emails are restriction to a few regions only and ca-central-1 isn't part of these:
https://docs.aws.amazon.com/general/latest/gr/ses.html

.github/workflows/merge_to_main_staging.yml
jzbahrai: i removed these from secrets and put them in terragrunt.hcl

aws/ses_receiving_emails/cloudwatch_alarms.tf
```diff
@@ -0,0 +1,34 @@
...
+  evaluation_periods  = "1"
+  metric_name         = aws_cloudwatch_log_metric_filter.ses_receiving_emails-500-errors-api.metric_transformation[0].name
+  namespace           = aws_cloudwatch_log_metric_filter.ses_receiving_emails-500-errors-api.metric_transformation[0].namespace
+  period              = "300"
+  statistic           = "Sum"
+  threshold           = 10
+  treat_missing_data  = "notBreaching"
+  alarm_actions       = [var.sns_alert_critical_arn]
+  ok_actions          = [var.sns_alert_critical_arn]
+}
```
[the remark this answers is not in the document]
jzbahrai: yes good call, commented out the critical alarm for now

aws/ses_receiving_emails/cloudwatch_logs.tf
jzbahrai: done

aws/ses_receiving_emails/lambda.tf
jzbahrai: ah ok, i just moved them to terragrunt.hcl dont think anything is sensitive here

aws/ses_receiving_emails/lambda.tf
```diff
@@ -0,0 +1,43 @@
+module "ses_receiving_emails" {
+  source                 = "github.com/cds-snc/terraform-modules?ref=v4.0.3//lambda"
```
jzbahrai, line 2: @patheard I was looking at the terraform-modules, based on @jimleroyer comment that this needs to run in us-east-1  - is there a place for me to setup the lambda in that region vs ca-central-1 (which I am assuming is what it defaults to)?

https://github.com/cds-snc/terraform-modules/blob/main/lambda/main.tf

patheard: Yup, you can have the lambda created in `us-east-1` by [passing in that provider to the module](https://developer.hashicorp.com/terraform/language/modules/develop/providers#passing-providers-explicitly) like so:
```hcl
module "ses_receiving_emails" {
  source    = "github.com/cds-snc/terraform-modules?ref=v4.0.3//lambda"
  providers = {
    aws = aws.us-east-1 # this is using the provider alias already defined in the project
  }
  
  # etc
}
```
And you're right about the default provider being `ca-central-1`.  Modules automatically [inherit the default provider config from their parent](https://developer.hashicorp.com/terraform/language/modules/develop/providers#implicit-provider-inheritance).

jzbahrai: done!

aws/ses_receiving_emails/lambda.tf
```diff
@@ -0,0 +1,43 @@
...
+
+  environment_variables = {
+    NOTIFY_SENDING_DOMAIN   = var.notify_sending_domain
+    SQS_REGION              = var.sqs_region
+    CELERY_QUEUE_PREFIX     = var.celery_queue_prefix
+    GC_NOTIFY_SERVICE_EMAIL = var.gc_notify_service_email
+  }
+}
+
+resource "aws_lambda_function_event_invoke_config" "ses_receiving_emails_invoke_config" {
```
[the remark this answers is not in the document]
jzbahrai: I have ideas around this, but lets discuss this when you come back.

aws/ses_receiving_emails/lambda.tf
patheard: Perfect, I believe you'll now also have to make the other resources in this file also get created in the same region by specifying the `aws.us-east-1` provider for them as well:
```hcl
resource "aws_lambda_function_event_invoke_config" "ses_receiving_emails_invoke_config" {
  provider = aws.us-east-1
}

# same for:
# aws_cloudwatch_event_target
# aws_cloudwatch_event_rule
# aws_lambda_permission
```

env/staging/ses_receiving_emails/terragrunt.hcl
```diff
@@ -0,0 +1,33 @@
+dependencies {
+  paths = ["../common"]
+}
+
+dependency "common" {
+  config_path = "../common"
+
+  # Configure mock outputs for the `validate` command that are returned when there are no outputs available (e.g the
+  # module hasn't been applied yet.
+  mock_outputs = {
```
patheard, line 10: Can you add the following two params as well please:
```hcl
mock_outputs_merge_with_state           = true
```
This will prevent mock values from accidentally being used with a `terraform apply` command and allow you to use mock values even if the Terraform state file exists for the module (it doesn't in this case, but it'll help if you ever add more mock values).

aws/ses_receiving_emails/variables.tf
```diff
@@ -0,0 +1,39 @@
+variable "billing_tag_value" {
+  type        = string
+  description = "Identifies the billing code."
+}
+
+variable "notify_sending_domain" {
+  sensitive   = true
```
patheard, line 7: Missed this on my last review 😅  Since this is now hardcoded in the `terragrunt.hcl` you can get rid of the `sensitive` param.

aws/ses_receiving_emails/variables.tf
```diff
@@ -0,0 +1,39 @@
...
+}
+
+variable "celery_queue_prefix" {
+  type        = string
+  description = "Celery queue prefix"
+}
+
+variable "gc_notify_service_email" {
+  type        = string
+  sensitive   = true
```
patheard, line 24: Same comment about removing `sensitive`.

patheard (approved):

PR #599, opened 2022-12-13
Add a new retention period to the celery log group

smcmurtry:
# Summary | Résumé

I am attempting to change the retention period of our celery logs from "indefinite" to 30 days. The purpose of this is to delete PII that was logged by sqlalchemy in the log group `/aws/containerinsights/notification-canada-ca-production-eks-cluster/application`. 

Pat has advised me that this log group was probably auto created, and does not exist in terraform yet. So I will need to:
> do a Terraform state import so that Terraform doesn’t try to create the log group, but instead only updates the retention period 

[Terraform state import for log group docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_log_group#import)

### DIFF
--- a/aws/eks/cloudwatch_log.tf
+++ b/aws/eks/cloudwatch_log.tf
@@ -7,6 +7,11 @@ resource "aws_cloudwatch_log_group" "notification-canada-ca-eks-cluster-logs" {
   retention_in_days = 14
 }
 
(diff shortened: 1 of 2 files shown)


smcmurtry:
After discussion offline, I decided to do this with click-ops instead.

PR #600, opened 2022-12-19
Run scan-files on the document-download s3 bucket

smcmurtry:
# Summary | Résumé

At @maxneuvians suggestion, I took a look at how `scan-files` was integrated in [share-files-securely](https://github.com/cds-snc/share-files-securely). I am copying some of the [terraform from that project]( https://github.com/cds-snc/share-files-securely/blob/main/terraform/file_scanning.tf#L1-L8) in this PR.

As I understand it, this will tell `scan-files` to start scanning the files uploaded by users that will be attached to emails. `scan-files` will then add some metadata to files in the s3 bucket, and we can implement something in `notification-api` to read that metadata and reject any files flagged as malicious. 

I will first deploy this to staging to test it out, and later make the change in production if everything is working as expected.

Check that the terragrunt plan is as expected.

### DIFF
--- a/aws/common/file_scanning.tf
+++ b/aws/common/file_scanning.tf
@@ -1,10 +1,8 @@
 module "s3_scan_objects" {
   source = "github.com/cds-snc/terraform-modules?ref=v3.0.9//S3_scan_object"
 
-  product_name = "gc-notify"
-  # just apply this to staging first to test it out
-  # after we QA, change to "notification-canada-ca-${var.env}-document-download"
-  s3_upload_bucket_name = "notification-canada-ca-staging-document-download"
+  product_name          = "gc-notify"
+  s3_upload_bucket_name = "notification-canada-ca-${var.env}-document-download"
 
   billing_tag_value = var.billing_tag_value
 }
--- a/aws/common/file_scanning.tf
+++ b/aws/common/file_scanning.tf
@@ -1,0 +1,10 @@
+module "s3_scan_objects" {
+  source = "github.com/cds-snc/terraform-modules?ref=v3.0.9//S3_scan_object"
+
+  product_name          = "gc-notify"
+  # just apply this to staging first to test it out
+  # after we QA, change to "notification-canada-ca-${var.env}-document-download"
+  s3_upload_bucket_name = "notification-canada-ca-staging-document-download"
+
+  billing_tag_value = var.billing_tag_value
+}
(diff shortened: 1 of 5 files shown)


aws/common/file_scanning.tf
```diff
@@ -0,0 +1,10 @@
+module "s3_scan_objects" {
+  source = "github.com/cds-snc/terraform-modules?ref=v3.0.9//S3_scan_object"
+
+  product_name = "gc-notify"
+  # just apply this to staging first to test it out
+  # after we QA, change to "notification-canada-ca-${var.env}-document-download"
+  s3_upload_bucket_name = "notification-canada-ca-staging-document-download"
```
maxneuvians, line 7: I would leave it `var.env` because if you applied this to the production account, it might fail because the account does not exist. The way terraform is applied between staging and production is using this file:

https://github.com/cds-snc/notification-terraform/blob/main/.github/workflows/infrastructure_version.txt

so unless you bump this version, the TF would never apply to production.

patheard:
I'm curious about why the `Staging: performance-test` plan is showing a change to the `resource "aws_security_group" "perf_test"` resource.   Without the PrivateLink and S3 gateway rules, I suspect the perf test will stop working.

smcmurtry:
@patheard Yes that seems weird to me too. Would putting the new `.tf` file  in its own folder reduce the number of changes?

patheard:
> @patheard Yes that seems weird to me too. Would putting the new `.tf` file in its own folder reduce the number of changes?

No, I don't think that would help.  I think Terraform is fighting with itself because of how that security group is defined.  It's using both an inline [`egress` rule block](https://github.com/cds-snc/notification-terraform/blob/feat/integrate-scan-files/aws/performance-test/security_group.tf#L6-L13) and [standalone rule definitions](https://github.com/cds-snc/notification-terraform/blob/feat/integrate-scan-files/aws/performance-test/security_group.tf#L18-L26).  To fix this, only one type of rule definition should be used, which would mean changing the `egress rule block` to a standalone rule def like so:

1. Delete the [`egress` block](https://github.com/cds-snc/notification-terraform/blob/feat/integrate-scan-files/aws/performance-test/security_group.tf#L6-L13).
2. Add a new standalone rule:
```hcl
resource "aws_security_group_rule" "perftest-egress-internet" {
  description              = "Egress to the internet from perftest"
  type                     = "egress"
  from_port                = 443
  to_port                  = 443
  protocol                 = "tcp"
  cidr_blocks              = ["0.0.0.0/0"]
  security_group_id        = aws_security_group.perf_test.id
}
```

patheard (approved):
🔬 📄

jimleroyer:
Good stuff @smcmurtry !

PR #601, opened 2022-12-19
Fix some billing tags for production infra that was tagged as staging

smcmurtry:
# Summary | Résumé

While working on https://github.com/cds-snc/notification-terraform/pull/600, I noticed that some of the billing tags for production infrastructure are being set to staging. Not a huge problem, but I assume this will make our cost reporting less accurate so thought I would fix this.


jimleroyer (approved):
Thanks @smcmurtry !

PR #609, opened 2023-01-17
Role name

jzbahrai:
# Summary | Résumé

Give the lambda function permissions to reach out to SQS

### DIFF
--- a/aws/ses_receiving_emails/lambda.tf
+++ b/aws/ses_receiving_emails/lambda.tf
@@ -42,29 +42,13 @@ resource "aws_lambda_permission" "ses_receiving_emails_allow_cloudwatch" {
   source_arn    = aws_cloudwatch_event_rule.ses_receiving_emails_testing.arn
 }
 
-resource "aws_iam_policy" "ses_recieving_emails_sqs_send" {
-  name        = "ses_recieving_email_sqs_send"
-  path        = "/"
-  description = "IAM policy for sending messages to SQS from Lambda"
-
-  policy = <<POLICY
-{
-    "Version": "2012-10-17",
-    "Statement": [
-        {
-            "Action": [
-                "sqs:Get*",
-                "sqs:SendMessage"
-            ],
-            "Effect": "Allow",
-            "Resource": "*"
-        }
+data "aws_iam_policy_document" "ses_recieving_emails_sqs_send" {
+  statement {
+    actions = [
+      "sqs:Get*",
+      "sqs:SendMessage"
     ]
-}
-POLICY
-}
-
-resource "aws_iam_role_policy_attachment" "ses_recieving_emails_sqs_send" {
-  role       = module.ses_receiving_emails.function_role_arn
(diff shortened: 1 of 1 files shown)


aws/ses_receiving_emails/lambda.tf
```diff
@@ -41,3 +41,30 @@ resource "aws_lambda_permission" "ses_receiving_emails_allow_cloudwatch" {
...
+        }
+    ]
+}
+POLICY
+}
+
+resource "aws_iam_role_policy_attachment" "ses_recieving_emails_sqs_send" {
+  role       = module.ses_receiving_emails.function_role_arn
+  policy_arn = aws_iam_policy.ses_recieving_emails_sqs_send.arn
+}
```
patheard, line 70: Instead of doing the policy attachment here (which would work as expected), you can pass the policy directly into the lambda module like so and it will do the attachment for you:
https://github.com/cds-snc/scan-files/blob/714391200db9bf30e31d3af1a1f571ccf13d1776/terragrunt/aws/api/lambda.tf#L30-L34

aws/ses_receiving_emails/lambda.tf
```diff
@@ -41,3 +41,30 @@ resource "aws_lambda_permission" "ses_receiving_emails_allow_cloudwatch" {
...
+                "sqs:Get*",
+                "sqs:SendMessage"
+            ],
+            "Effect": "Allow",
+            "Resource": "*"
+        }
+    ]
+}
+POLICY
+}
```
patheard, line 65: Could you change this to define the policy using a `data` element.  That way we'll get static syntax checking before we do the TF apply.

Here's an example of [creating the policy using the `data` element](https://github.com/cds-snc/scan-files/blob/714391200db9bf30e31d3af1a1f571ccf13d1776/terragrunt/aws/api/iam.tf#L172-L181).  And this is how you [define the `data` element](https://github.com/cds-snc/scan-files/blob/714391200db9bf30e31d3af1a1f571ccf13d1776/terragrunt/aws/api/iam.tf#L201-L226).

aws/ses_receiving_emails/lambda.tf
```diff
@@ -41,3 +41,25 @@ resource "aws_lambda_permission" "ses_receiving_emails_allow_cloudwatch" {
...
+resource "aws_iam_policy" "ses_recieving_emails_sqs_send" {
+  name        = "ses_recieving_email_sqs_send"
+  description = "IAM policy for sending messages to SQS from Lambda"
+  policy      = data.aws_iam_policy_document.ses_recieving_emails_sqs_send.json
+
+  tags = {
+    CostCentre = var.billing_tag_value
+    Terraform  = true
+  }
+}
```
patheard, line 65: I realize my confusing comments didn't help, but you can get rid of this element.

Then you just need to specify the new policy document on the lambda module like so:

```hcl
module "ses_receiving_emails" {
   ...
   policies = [
      data.aws_iam_policy_document.ses_recieving_emails_sqs_send.json
   ]
   ...
}

```

patheard (approved):

PR #614, opened 2023-01-23
Fix invocation of ses_receiving_lambda

jzbahrai:
# Summary | Résumé

Fix invocation of ses_receiving_lambda.

### DIFF
--- a/aws/dns/ses.tf
+++ b/aws/dns/ses.tf
@@ -159,7 +159,7 @@ resource "aws_ses_domain_mail_from" "custom_sending_domains" {
 }
 
 
-### SES recieving emails lambda image
+### SES receiving emails lambda image
 
 resource "aws_ses_receipt_rule" "ses_receiving_emails_inbound-to-lambda-arn" {
   provider = aws.us-east-1
--- a/aws/dns/ses.tf
+++ b/aws/dns/ses.tf
@@ -157,3 +157,22 @@ resource "aws_ses_domain_mail_from" "custom_sending_domains" {
   domain           = aws_ses_domain_identity.custom_sending_domains[each.value].domain
   mail_from_domain = "bounce.${aws_ses_domain_identity.custom_sending_domains[each.value].domain}"
 }
+
+
+### SES recieving emails lambda image
+
+resource "aws_ses_receipt_rule" "ses_receiving_emails_inbound-to-lambda-arn" {
+  provider = aws.us-east-1
+
+  name          = "ses_receiving_emails_inbound-to-lambda"
+  rule_set_name = aws_ses_receipt_rule_set.main.rule_set_name
+  recipients    = [var.domain]
+  enabled       = true
+  scan_enabled  = true
+
+  lambda_action {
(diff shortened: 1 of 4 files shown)


aws/dns/ses.tf
```diff
@@ -157,3 +157,22 @@ resource "aws_ses_domain_mail_from" "custom_sending_domains" {
   domain           = aws_ses_domain_identity.custom_sending_domains[each.value].domain
   mail_from_domain = "bounce.${aws_ses_domain_identity.custom_sending_domains[each.value].domain}"
 }
+
+
+### SES recieving emails lambda image
```
jimleroyer, line 162:
```suggestion
### SES receiving emails lambda image
```

aws/dns/ses.tf
```diff
@@ -157,3 +157,22 @@ resource "aws_ses_domain_mail_from" "custom_sending_domains" {
...
+  recipients    = [var.domain]
+  enabled       = true
+  scan_enabled  = true
+
+  lambda_action {
+    function_arn    = var.lambda_ses_receiving_emails_image_arn
+    invocation_type = "Event"
+    position        = 1
+  }
+}
```
jimleroyer, line 178: While we add this new receipt rule, do we want to delete the existing one to the other lambda? i.e. aws_ses_receipt_rule / inbound-to-lambda? We could leave it to test the new lambda while the other one is still functional, but we'd be in a state where potentially two lambdas can send no reply emails. How do we want to proceed?

patheard: That's a good call - I'd recommend changing the name of this receiving rule so you can test both and don't hit conflicts trying to create a duplicate rule.

jzbahrai: @jimleroyer I wanted both lambda to run to see them working for now. I will do the deletes after.

In regards to the rule set - I was reading this: https://docs.aws.amazon.com/ses/latest/dg/receiving-email-concepts.html 

I think this part remains the same: https://github.com/cds-snc/notification-terraform/pull/614/files#diff-ed0d44114802d77c0c61d116fd96c29d08470a5992e2874598c504d3325d98f2R48-R58

but then the name of this: https://github.com/cds-snc/notification-terraform/pull/614/files#diff-ed0d44114802d77c0c61d116fd96c29d08470a5992e2874598c504d3325d98f2R48-R58

needs to be different from the receipt_rule https://github.com/cds-snc/notification-terraform/pull/614/files#diff-ed0d44114802d77c0c61d116fd96c29d08470a5992e2874598c504d3325d98f2R60

so I think we should be okay? I am not sure, and would like to test this on staging though.

jimleroyer: Is my assumption wrong that both lambdas will run and we will have 2 no reply emails to be sent? If yes, either we'd need to test this quick in staging and bring necessary fix/changes once deployed, or have only one enabled. I'd prefer if we minimize the time window in staging where we test this because that would block our release pipeline and would be in the way of an emergency fix.

jimleroyer (approved):
Let's test this out in staging. 🚀

PR #617, opened 2023-01-23
Add a new elasticache cluster

jzbahrai:
# Summary | Résumé

Add a new redis cluster with 1 node, 2 replicas. This is the first step of the migration plan to move redis to a new multi-AZ cluster.

### DIFF
--- a/env/production/elasticache/terragrunt.hcl
+++ b/env/production/elasticache/terragrunt.hcl
@@ -21,7 +21,7 @@ include {
 inputs = {
   eks_cluster_securitygroup = dependency.eks.outputs.eks-cluster-securitygroup
   elasticache_node_count    = 1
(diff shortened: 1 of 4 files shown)


env/production/elasticache/terragrunt.hcl
```diff
@@ -21,6 +21,7 @@ include {
 inputs = {
   eks_cluster_securitygroup = dependency.eks.outputs.eks-cluster-securitygroup
   elasticache_node_count    = 1
+  elasticache_node_number_cache_clusters = 3
```
sastels, line 24: Can you reformat so all the "="s are in a line (VScode does this automatically for terraform files but we need to do manually for terragrunt)

sastels (approved):
LGTM except for the formatting!

PR #624, opened 2023-01-31
Remove old redis cluster and add new metrics

jzbahrai:
# Summary | Résumé

We want to remove the old redis cluster and set up the cloudwatch alarms to use the new redis cluster

### DIFF
--- a/aws/elasticache/cloudwatch_alarms.tf
+++ b/aws/elasticache/cloudwatch_alarms.tf
@@ -17,25 +17,6 @@ resource "aws_cloudwatch_metric_alarm" "redis-elasticache-medium-cpu-warning" {
   }
 }
 
-# We created a node-4 which doesnt have the same naming convention as the rest of the nodes,
-# In order to rename the node we would have to delete the old one on the live cluster, we are going to chose not to do this
-resource "aws_cloudwatch_metric_alarm" "redis-elasticache-medium-cpu-warning-node4" {
-  alarm_name          = "redis-elasticache-medium-cpu-warning-CacheClustermultiaz-group-replicaCPUUtilization"
-  alarm_description   = "Average CPU of Redis ElastiCache >= 50% during 1 minute"
-  comparison_operator = "GreaterThanOrEqualToThreshold"
-  evaluation_periods  = "1"
-  metric_name         = "CPUUtilization"
-  namespace           = "AWS/ElastiCache"
-  period              = "60"
-  statistic           = "Average"
-  threshold           = 50
-  alarm_actions       = [var.sns_alert_warning_arn]
-  treat_missing_data  = "breaching"
-  dimensions = {
-    CacheClusterId = aws_elasticache_cluster.notification-cluster-cache-multiaz-group-replica.cluster_id
-  }
-}
-
 resource "aws_cloudwatch_metric_alarm" "redis-elasticache-high-cpu-warning" {
   count               = var.elasticache_node_number_cache_clusters
   alarm_name          = "redis-elasticache-high-cpu-warning-CacheCluster00${count.index + 1}CPUUtilization"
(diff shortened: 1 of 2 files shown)


aws/elasticache/cloudwatch_alarms.tf
```diff
@@ -11,7 +11,7 @@ resource "aws_cloudwatch_metric_alarm" "redis-elasticache-medium-cpu-warning" {
   alarm_actions       = [var.sns_alert_warning_arn]
   treat_missing_data  = "breaching"
   dimensions = {
-    CacheClusterId = aws_elasticache_cluster.notification-cluster-cache.cluster_id
+    ReplicationGroupId = aws_elasticache_replication_group.notification-cluster-cache-multiaz-group.id
```
patheard, line 14: From what I can see, these metrics don't exist under the `RepicationGroupId` and are still under the `CacheClusterId`.  The gotcha is that they're now split up by node within the cluster.  

It looks like you can get an aggregate of CPUUtilization across all clusters in the account if you remove the `dimensions` but I haven't tried it in Terraform (this was just from playing in the AWS console).

aws/elasticache/cloudwatch_alarms.tf
```diff
@@ -44,7 +44,7 @@ resource "aws_cloudwatch_metric_alarm" "redis-elasticache-high-db-memory-warning
   threshold           = 60
   alarm_actions       = [var.sns_alert_warning_arn]
   dimensions = {
-    CacheClusterId = aws_elasticache_cluster.notification-cluster-cache.cluster_id
+    ReplicationGroupId = aws_elasticache_replication_group.notification-cluster-cache-multiaz-group.id
```
patheard, line 47: Same comment - I don't think this metric dimension exists.

aws/elasticache/cloudwatch_alarms.tf
jzbahrai: thanks @patheard 
I tried using this approach instead: https://github.com/terraform-community-modules/tf_aws_elasticache_redis/blob/master/cloudwatch.tf#L21

unfortunately the 4th node has a different name, so I had to make its own cloudwatch alarm for it.

aws/elasticache/cloudwatch_alarms.tf
```diff
@@ -10,13 +11,49 @@ resource "aws_cloudwatch_metric_alarm" "redis-elasticache-medium-cpu-warning" {
...
   alarm_actions       = [var.sns_alert_warning_arn]
   treat_missing_data  = "breaching"
+
   dimensions = {
-    CacheClusterId = aws_elasticache_cluster.notification-cluster-cache.cluster_id
+    CacheClusterId = "${aws_elasticache_replication_group.notification-cluster-cache-multiaz-group.id}-00${count.index + 1}"
+  }
+}
+
+resource "aws_cloudwatch_metric_alarm" "redis-elasticache-medium-cpu-warning-node4" {
```
jimleroyer, line 20: This is unfortunate we need to add one more resource because it has a different name. Do you know why it led to this naming? would that be possible to align that node with the others? Else let's put a comment to explain the situation in the TF file please. 🙏

aws/elasticache/cloudwatch_alarms.tf
```diff
@@ -10,13 +11,49 @@ resource "aws_cloudwatch_metric_alarm" "redis-elasticache-medium-cpu-warning" {
...
+  statistic           = "Average"
+  threshold           = 70
+  alarm_actions       = [var.sns_alert_warning_arn]
+  treat_missing_data  = "breaching"
+  dimensions = {
+    CacheClusterId = "${aws_elasticache_replication_group.notification-cluster-cache-multiaz-group.id}-00${count.index + 1}"
+  }
+}
+
+resource "aws_cloudwatch_metric_alarm" "redis-elasticache-high-cpu-warning-node4" {
```
jimleroyer, line 55: Same comment as before would apply here.

aws/elasticache/cloudwatch_alarms.tf
```diff
@@ -28,12 +65,29 @@ resource "aws_cloudwatch_metric_alarm" "redis-elasticache-high-cpu-warning" {
...
+  period              = "60"
+  statistic           = "Average"
+  threshold           = 60
+  alarm_actions       = [var.sns_alert_warning_arn]
+  dimensions = {
+    CacheClusterId = "${aws_elasticache_replication_group.notification-cluster-cache-multiaz-group.id}-00${count.index + 1}"
+  }
+}
+
+resource "aws_cloudwatch_metric_alarm" "redis-elasticache-high-db-memory-warning-node4" {
```
jimleroyer, line 89: Same comment, yaddi yadda..

aws/elasticache/cloudwatch_alarms.tf
```diff
@@ -44,12 +98,29 @@ resource "aws_cloudwatch_metric_alarm" "redis-elasticache-high-db-memory-warning
...
+  period              = "60"
+  statistic           = "Average"
+  threshold           = 1000
+  alarm_actions       = [var.sns_alert_warning_arn]
+  dimensions = {
+    CacheClusterId = "${aws_elasticache_replication_group.notification-cluster-cache-multiaz-group.id}-00${count.index + 1}"
+  }
+}
+
+resource "aws_cloudwatch_metric_alarm" "redis-elasticache-high-connection-warning-node4" {
```
jimleroyer, line 122: 👯

aws/elasticache/elasticache.tf
```diff
@@ -7,30 +7,6 @@ resource "aws_elasticache_subnet_group" "notification-canada-ca-cache-subnet" {
   subnet_ids = var.vpc_private_subnets
 }
 
-resource "aws_elasticache_cluster" "notification-cluster-cache" {
```
jimleroyer, line 10: Veury noice! 💥

aws/elasticache/cloudwatch_alarms.tf
patheard: Makes sense.  I think there's a risk my suggestion of the aggregate metrics would hide cases where one of the nodes was struggling and the others were fine.

aws/elasticache/cloudwatch_alarms.tf
jzbahrai: added a comment

jimleroyer (approved):
Let's test this out in staging. 🚀

PR #625, opened 2023-02-02
Create S3 bucket for file attachments for scan-files to run on

smcmurtry:
# Summary | Résumé

This PR creates a duplicate S3 bucket that will hold file attachments uploaded to the document download API. File retention periods are the same as the existing bucket. I am also enabling the `S3_scan_object` module to run on this bucket.

Verify that the terraform plans look correct and that the code changes make sense.


aws/common/file_scanning.tf
```diff
@@ -0,0 +1,8 @@
+module "s3_scan_objects" {
+  source = "github.com/cds-snc/terraform-modules?ref=v3.0.9//S3_scan_object"
```
patheard, line 2: You should be able to use the latest release of this module:
```suggestion
  source = "github.com/cds-snc/terraform-modules?ref=v5.0.0//S3_scan_object"
```

aws/common/s3.tf
```diff
@@ -249,6 +249,51 @@ resource "aws_s3_bucket" "document_bucket" {
...
+  #tfsec:ignore:AWS077 - Versioning is not enabled
+  logging {
+    target_bucket = aws_s3_bucket.document_bucket_logs.bucket
+  }
+
+  tags = {
+    CostCenter = "notification-canada-ca-${var.env}"
+  }
+}
+
```
patheard, line 296: If you wanted, you could use the CDS `S3` module to create the bucket like so:
```suggestion
module "scan_files_document_bucket" {
  source      = "github.com/cds-snc/terraform-modules?ref=v3.0.20//S3"
  bucket_name = "notification-canada-ca-${var.env}-document-download-scan-files"
  lifecycle_rule = [
    {
      id      = "tf-s3-lifecycle-linked-files"
      enabled = true
      expiration {
        days = 7
      }
    },
    {
      id      = "tf-s3-lifecycle-attached-files"
      enabled = true
      prefix  = "tmp/"
      expiration {
        days = 3
      }
    }
  ]
  billing_tag_value = "notification-canada-ca-${var.env}"
  logging = {
    target_bucket = aws_s3_bucket.document_bucket_logs.bucket
  }
}
```

aws/common/s3.tf
```diff
@@ -258,6 +303,15 @@ resource "aws_s3_bucket_public_access_block" "document_bucket" {
...
 
+resource "aws_s3_bucket_public_access_block" "scan_files_document_bucket" {
+  bucket = aws_s3_bucket.scan_files_document_bucket.id
+
+  block_public_acls       = true
+  block_public_policy     = true
+  ignore_public_acls      = true
+  restrict_public_buckets = true
+}
+
```
patheard, line 314: If you use the CDS `S3` module, you can get rid of this resource as well since the module creates it.

patheard (commented):
Looks good Steve!  Just had a minor comment about the Scan Object version and the option to use our `S3` module (but what you've got should also work just fine).

aws/common/s3.tf
```diff
@@ -249,6 +249,51 @@ resource "aws_s3_bucket" "document_bucket" {
...
 
+resource "aws_s3_bucket" "scan_files_document_bucket" {
+  bucket = "notification-canada-ca-${var.env}-document-download-scan-files"
+  acl    = "private"
+
+  server_side_encryption_configuration {
+    rule {
+      apply_server_side_encryption_by_default {
+        sse_algorithm = "AES256"
+      }
```
jimleroyer, line 260: @patheard Steve and I were wondering what kind of encryption you wanted in there. Do you want encryption with an key stored in KMS and that will be cross-account compatible?

aws/common/s3.tf
```diff
@@ -249,6 +249,51 @@ resource "aws_s3_bucket" "document_bucket" {
   }
 }
 
+resource "aws_s3_bucket" "scan_files_document_bucket" {
+  bucket = "notification-canada-ca-${var.env}-document-download-scan-files"
```
jimleroyer, line 253: Good name following the naming conventions. 👌

aws/common/s3.tf
```diff
@@ -249,6 +249,51 @@ resource "aws_s3_bucket" "document_bucket" {
...
+    }
+  }
+
+  # expire linked files after 7 days
+  lifecycle_rule {
+    id      = "tf-s3-lifecycle-linked-files"
+    enabled = true
+
+    expiration {
+      days = 7
```
jimleroyer, line 270: Can we be more aggressive on the files expiration with this bucket? We just need these for the time for scan files to process. @patheard Do you know the expected maximum time for the scan file service to process the files?

aws/common/s3.tf
jimleroyer: Is encryption automatically handled with this module? If yes, default encryption? i.e. no KMS key?

aws/common/s3.tf
jimleroyer: That is good selling over there. 👍

aws/common/s3.tf
patheard: For sure - the scan should complete within 15 minutes at most, but more likely it's under 1 minute.

aws/common/s3.tf
patheard: Yup, exactly so - it defaults to `AES256` if a KMS key isn't specified:
https://github.com/cds-snc/terraform-modules/blob/8c3bdffea8b6f8dc8244e1a5f1a8e0141cd4fd28/S3/main.tf#L66-L75

aws/common/s3.tf
patheard: `AES256` is totally fine to use here.

aws/common/s3.tf
jimleroyer: If security team is good with it, we are good with it. We should have @ascheid looped in for reviewing and the [ADR a well](https://github.com/cds-snc/notification-adr/pull/30).

aws/common/s3.tf
smcmurtry: removed!

ascheid:
Reviewed the ADR and had a chat with @smcmurtry and @jimleroyer. As long as we keep at least the sec controls we have for the normal S3 bucket, I don't see any issues. @jimleroyer mentioned that we would have tighter access policies for the files. If the files is only going to be accessed by scan-files, I suggested that we only use the tag scan-files is going to write.

aws/common/s3.tf
smcmurtry: I've switched the expiration to 1 day for both buckets

aws/common/s3.tf
```diff
@@ -249,6 +249,51 @@ resource "aws_s3_bucket" "document_bucket" {
...
+  # See https://github.com/cds-snc/notification-document-download-api
+  lifecycle_rule {
+    id      = "tf-s3-lifecycle-attached-files"
+    enabled = true
+    prefix  = "tmp/"
+
+    expiration {
+      days = 1
+    }
+  }
```
patheard, line 285: Since this now has the same expiration timeframe as the other `lifecycle_rule`, can we just remove it?

smcmurtry: yes good point!

patheard (approved):

PR #640, opened 2023-02-07
Added permissions for refactored SQS to email callback 

jzbahrai:
# Summary | Résumé

[ZenHub task](https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/1066)

I added the permissions so that sns can execute this lambda, but we also need the lambda to subscribe to the correct sns topic.

I also added some print statements in the current lambda so that I can see what is happening in the logs. This way I can do a comparison when the new lambda runs.


sastels (approved):
🚀

andrewleith (approved):
LGTM

PR #641, opened 2023-02-08
SNS subscription needs to be function arn and not name

jzbahrai:
# Summary | Résumé

Made an error and put function_name instead of function_arn for endpoint in sns subscription


andrewleith (approved):
👍

PR #642, opened 2023-02-08
Add permissions for sqs

jzbahrai:
# Summary | Résumé

The lambda is failing as we can't read from SQS. This is the same permission that was added here:
https://github.com/cds-snc/notification-terraform/blob/main/aws/ses_receiving_emails/lambda.tf#L18-L32

so this is similar but for the next lambda

Screenshot 2023-02-08 at 11 45 23 AM

### DIFF
--- a/aws/ses_to_sqs_email_callbacks/lambda.tf
+++ b/aws/ses_to_sqs_email_callbacks/lambda.tf
@@ -8,6 +8,20 @@ module "ses_to_sqs_email_callbacks" {
   timeout                = 60
   memory                 = 1024
 
+  policies = [
+    data.aws_iam_policy_document.ses_to_sqs_email_callbacks.json
+  ]
+}
+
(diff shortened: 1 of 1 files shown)


sastels (approved):

aws/ses_to_sqs_email_callbacks/lambda.tf
```diff
@@ -8,6 +8,20 @@ module "ses_to_sqs_email_callbacks" {
...
+}
+
+data "aws_iam_policy_document" "ses_to_sqs_email_callbacks" {
+  statement {
+    actions = [
+      "sqs:Get*",
+      "sqs:SendMessage"
+    ]
+    effect    = "Allow"
+    resources = ["*"]
```
jimleroyer, line 23: Can we restrict the resource to the lambda?

PR #643, opened 2023-02-08
Remove old lambda invocation and move the image to production

jzbahrai:
# Summary | Résumé

Switch off the old lambda, and switch on the new one

You can see that both lambdas are currently running in staging.
I am testing this by looking at the log output of both the image lambda (ses_to_sqs_email_callbacks) vs the file lambda (ses-to-sqs-email-callbacks)

File log:
Screenshot 2023-02-08 at 1 25 48 PM

Image log
Screenshot 2023-02-08 at 1 25 11 PM

Once this PR is merged in, the file lambda will not be invoked and only the image lambda will run. 
This PR once tagged is what will allow us to deploy this code to production

### DIFF
--- a/.github/workflows/terragrunt_plan_production.yml
+++ b/.github/workflows/terragrunt_plan_production.yml
@@ -75,6 +75,15 @@ jobs:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
           terragrunt: "true"
 
+      - name: Terragrunt plan ses_to_sqs_email_callbacks
+        uses: cds-snc/terraform-plan@v1
+        with:
+          directory: "env/production/ses_to_sqs_email_callbacks"
+          comment-delete: "true"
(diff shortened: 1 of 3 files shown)


.github/workflows/terragrunt_plan_production.yml
```diff
@@ -75,6 +75,15 @@ jobs:
...
 
+      - name: Terragrunt plan ses_to_sqs_email_callbacks
+        uses: cds-snc/terraform-plan@v1
+        with:
+          directory: "env/production/ses_to_sqs_email_callbacks"
+          comment-delete: "true"
+          comment-title: "Production: ses_to_sqs_email_callbacks"
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+          terragrunt: "true"
+
```
jimleroyer, line 86: Did we test in staging if that works before applying the actions to the production env?

jimleroyer (approved):
LGTM

PR #654, opened 2023-02-14
Fix scope of resources using arn

jzbahrai:
# Summary | Résumé

I thought I could put the name of the resource for IAM permissions, but it requires an ARN:
```
Plan: 0 to add, 1 to change, 0 to destroy.
module.ses_receiving_emails.aws_iam_policy.policies[0]: Modifying... [id=arn:aws:iam::239043911459:policy/ses_receiving_emails-0]
╷
│ Error: error updating IAM policy arn:aws:iam::239043911459:policy/ses_receiving_emails-0: MalformedPolicyDocument: Resource eks-notification-canada-ca/notify-internal-tasks must be in ARN format or "*".
│ 	status code: 400, request id: ce07e1c3-f7e5-498b-b1e5-142b5c153101
│ 
│   with module.ses_receiving_emails.aws_iam_policy.policies[0],
│   on .terraform/modules/ses_receiving_emails/lambda/user_policies.tf line 2, in resource "aws_iam_policy" "policies":
│    2: resource "aws_iam_policy" "policies" {
│ 
╵
time=2023-02-14T18:37:36Z level=error msg=1 error occurred:
	* exit status 1
```

Added the ARNs.

# NOTE:

The sqs queues were clickopsed. We are just adding them as resources so we have access to them in TF. I do not know if this is the right way to do this but will check after the merge.

### DIFF
--- a/aws/common/outputs.tf
+++ b/aws/common/outputs.tf
@@ -113,3 +113,7 @@ output "sns_deliveries_failures_us_west_2_arn" {
 output "sns_deliveries_failures_us_west_2_name" {
   value = aws_cloudwatch_log_group.sns_deliveries_failures_us_west_2.name
 }
+
+output "sqs_notify_internal_tasks_arn" {
+  value = aws_sqs_queue.notify-internal-tasks.arn
+}
--- a/aws/common/outputs.tf
+++ b/aws/common/outputs.tf
@@ -113,3 +113,11 @@ output "sns_deliveries_failures_us_west_2_arn" {
 output "sns_deliveries_failures_us_west_2_name" {
   value = aws_cloudwatch_log_group.sns_deliveries_failures_us_west_2.name
 }
+
+output "sqs_notify_internal_tasks_arn" {
+  value = aws_sqs_queue.notify-internal-tasks.arn
+}
+
+output "sqs_delivery_receipts_arn" {
+  value = aws_sqs_queue.sqs_delivery_receipts.arn
+}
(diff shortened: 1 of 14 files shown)


aws/common/sqs.tf
```diff
@@ -21,3 +21,13 @@ resource "aws_sqs_queue" "bulk_db_tasks_queue" {
   # tfsec:ignore:AWS015 - Queues should be encrypted with customer managed KMS keys
   # AWS managed encryption is good enough for us
 }
+
+resource "aws_sqs_queue" "notify-internal-tasks" {
```
jzbahrai, line 25: These resources already exist, I am unsure if this is the correct way to add them to tf

patheard: For the resources that already exist, there are a few different options.

1.  If you want to manage them with Terraform going forward, this is correct and we'll just have to do a TF state import to stop Terraform from trying to create them.
2. If you just need the ARNs, you can look them up using a [`data` element](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/sqs_queue) like so:
```hcl
data "aws_sqs_queue" "notify-internal-tasks" {
  name = "${var.celery_queue_prefix}notify-internal-tasks"
}
```
And then you can get the ARN with `data.aws_sqs_queue.notify-internal-tasks.arn`.

jimleroyer: These queues were initially created and now managed by Celery, so option #2 is better.

aws/common/outputs.tf
```diff
@@ -113,3 +113,7 @@ output "sns_deliveries_failures_us_west_2_arn" {
 output "sns_deliveries_failures_us_west_2_name" {
   value = aws_cloudwatch_log_group.sns_deliveries_failures_us_west_2.name
 }
+
+output "sqs_notify_internal_tasks_arn" {
+  value = data.aws_sqs_queue.notify-internal-tasks.arn
+}
```
jimleroyer, line 119: 🎯

jimleroyer (approved):
LGTM!

PR #659, opened 2023-02-21
Add alarms for malware detection

andrewleith:
# Summary | Résumé

This PR adds a warning and a critical alarm for when malware is detected.

- Warning occurs when 1 event of malware detection happens in a 1 minute period
- Critical occurs when 10 events of malware detection happens in a 1 minute period

### DIFF
--- a/aws/eks/cloudwatch_log.tf
+++ b/aws/eks/cloudwatch_log.tf
@@ -36,7 +36,7 @@ resource "aws_cloudwatch_log_metric_filter" "celery-error" {
 
 resource "aws_cloudwatch_log_metric_filter" "malware-detected" {
   name           = "malware-detected"
-  pattern        = "Malicious content detected"
+  pattern        = "Malicious content detected! Download and attachment failed"
   log_group_name = local.eks_application_log_group
 
   metric_transformation {
--- a/aws/eks/cloudwatch_log.tf
+++ b/aws/eks/cloudwatch_log.tf
@@ -36,11 +36,11 @@ resource "aws_cloudwatch_log_metric_filter" "celery-error" {
 
 resource "aws_cloudwatch_log_metric_filter" "malware-detected" {
   name           = "malware-detected"
-  pattern        = "?\"ERROR/Worker\" ?\"ERROR/ForkPoolWorker\" ?\"WorkerLostError\""
+  pattern        = "Malicious content detected"
   log_group_name = local.eks_application_log_group
 
   metric_transformation {
-    name      = "celery-error"
+    name      = "malware-detected"
     namespace = "LogMetrics"
     value     = "1"
   }
--- a/aws/eks/cloudwatch_log.tf
+++ b/aws/eks/cloudwatch_log.tf
@@ -33,3 +33,15 @@ resource "aws_cloudwatch_log_metric_filter" "celery-error" {
     value     = "1"
   }
 }
+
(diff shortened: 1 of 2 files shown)


aws/eks/cloudwatch_log.tf
```diff
@@ -33,3 +33,15 @@ resource "aws_cloudwatch_log_metric_filter" "celery-error" {
     value     = "1"
   }
 }
+
+resource "aws_cloudwatch_log_metric_filter" "malware-detected" {
+  name           = "malware-detected"
+  pattern        = "?\"ERROR/Worker\" ?\"ERROR/ForkPoolWorker\" ?\"WorkerLostError\""
```
smcmurtry, line 39: This is the same pattern as the celery error above: https://github.com/cds-snc/notification-terraform/pull/659/files#diff-40738109cc9f7aa8ec6a2a1f292fa717335630a1697dc151af763b63573ab682R27

I think this pattern needs to match the malware detected log right? Or am I missing something?

andrewleith: Whoops, thank you, must have over-written my change here!

andrewleith: Wouldn't have gotten too far like that! 😅 Just pushed a commit to do this properly!

jzbahrai: Should we make the alarms less slack? Like warning - 10 files in a 5 minute period
critical - 20 files in a 5 minute period?

I am unsure about the 1 file in 1 minute warning. Though the 10 files in 1 minute can remain for critical

aws/eks/cloudwatch_log.tf
```diff
@@ -33,3 +33,15 @@ resource "aws_cloudwatch_log_metric_filter" "celery-error" {
     value     = "1"
   }
 }
+
+resource "aws_cloudwatch_log_metric_filter" "malware-detected" {
+  name           = "malware-detected"
+  pattern        = "Malicious content detected"
```
smcmurtry, line 39:
```suggestion
  pattern        = "Malicious content detected! Download and attachment failed"
```

Could we target this to either just the celery log or just the dd-api log? I think this pattern will match on both logs and then we will be double counting. My suggested code just matches the log from Celery.

andrewleith: Sure! I wasn't sure how to target the logs (it just seemed to be looking at the entire EKS log group).  In my clickops testing in staging it didn't double count it, but I like your idea of making it more explicit.  I will make this update!

andrewleith: Done.

aws/eks/cloudwatch_log.tf
andrewleith: Good question @jzbahrai. I think the idea here is that this is rare enough to warrant knowing about it even when there is just 1.  But maybe @smcmurtry you could give your thoughts?

jimleroyer (approved):
LGTM! 

Will we add alarms (could be just warnings) if we got timeout from the malware scanning? Steve is almost getting these changes in.

andrewleith:
> LGTM!
> 
> Will we add alarms (could be just warnings) if we got timeout from the malware scanning? Steve is almost getting these changes in.

Certainly can.  Once that's done I can do another PR to add those.

PR #660, opened 2023-02-21
Feat/alarms for malware 2

andrewleith:
# Summary | Résumé
This PR updates the log metric filter pattern for malware detection to escape it properly.


jimleroyer (approved):
Let's try this out.

PR #662, opened 2023-02-22
Change celery warning alert to 5

jzbahrai:
# Summary | Résumé

Change celery warning alert to 5 in 1 minute vs 1 in 1 minute


jimleroyer:
We had a talk on Slack on how to handle individual failures. We want to target specific failures because these can have different impact. We do not want to treat all celery errors as the same (and silence more serious issues). Silencing sparse SSL connection errors for example to reduce the noisy alerts could be a possibility.

PR #665, opened 2023-02-22
Remove hosts restrictions and consolidate all invalid paths under one rule (to ring them all)

jimleroyer:
# Summary | Résumé

Because we want to catch fuzzy attacks bypassing our rules by hitting directly our EC2 hostnames (or IPs), these changes remove the hosts restrictions on invalid paths. **This means that all traffic going through our Kubernetes cluster and nodes will need to conform with these paths.** This is a test; this has some risk and that might not get deployed to production. We would merge this to staging after an explicit go ahead followed by tests, and a potential revert.

The admin and document download api rules are now consolidated into one rule. These rules could be separate before because they would apply on only one host (dd-api or admin). But now as there are no such filter on a sub-selection of rules, all rules apply globally on incoming traffic, hence all regular expressions of invalid paths need to be evaluated at the same time so that allowed exceptions can get through.

* Cross fingers. 🤞 
* Launch smoke tests on API.
* Test the admin with regular actions.
* Check WAF logs to see if we block expected traffic.
* Test various tools we use every day.
* Wait a day to check if anyone is screaming.


aws/eks/waf.tf
```diff
@@ -413,7 +330,7 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
 
     visibility_config {
       cloudwatch_metrics_enabled = true
-      metric_name                = "admin_invalid_path"
```
sastels, line 416: "invalid_paths" for the metric name? the metric is going to count the number of times the rule applies (ie there's a bad path)

sastels:
We should either scrap api-k8s or add the api regexs to this.

aws/eks/waf.tf
jimleroyer: Yeah I caught just before  your comment and that was changed in [there](https://github.com/cds-snc/notification-terraform/pull/665/commits/281fbc094c49e75bb2999d30d86ead50f225e6b4).

jimleroyer: I went with valid_paths instead though, considering @andrewleith suggestion on Slack.

sastels (approved):
LGTM! let's try it out!

aws/common/outputs.tf
```diff
@@ -58,6 +58,18 @@ output "ip_blocklist_arn" {
...
+}
+
+output "re_admin_arn" {
+  value = aws_wafv2_regex_pattern_set.re_admin.arn
+}
+
+output "re_document_download_arn" {
+  value = aws_wafv2_regex_pattern_set.re_document_download.arn
+}
+
```
patheard, line 72: A minor comment - could you add a description around what these are and how they're used?  Just for future us.

env/production/eks/terragrunt.hcl
```diff
@@ -14,9 +14,9 @@ dependency "common" {
   mock_outputs_allowed_terraform_commands = ["init", "fmt", "validate", "plan", "show"]
   mock_outputs_merge_with_state           = true
   mock_outputs = {
```
patheard, line 16: Could you add the three new regex ARN outputs as `mock_outputs` here.

env/production/lambda-api/terragrunt.hcl
```diff
@@ -51,7 +51,7 @@ dependency "dns" {
   # module hasn't been applied yet.
   mock_outputs_allowed_terraform_commands = ["init", "fmt", "validate", "plan", "show"]
   mock_outputs = {
```
patheard, line 53: Same comment about add `re_api_arn` as a `mock_outputs` here.

env/staging/lambda-api/terragrunt.hcl
```diff
@@ -22,6 +22,9 @@ dependency "common" {
     s3_bucket_csv_upload_bucket_arn = ""
     firehose_waf_logs_iam_role_arn  = ""
     ip_blocklist_arn                = ""
+    re_admin_arn                    = ""
+    re_api_arn                      = ""
+    re_document_download_arn        = ""
```
patheard, line 27: You'll only need `re_api_arn` as a mock output for this module.

env/production/eks/terragrunt.hcl
jimleroyer: Done..

patheard (approved):

PR #674, opened 2023-03-01
make bulk_send key inactive

sastels:
# Summary | Résumé

Don't need this anymore. Switching to inactive for now. Likely deleting the whole thing at a later point.


jimleroyer (approved):
:shipit:

PR #675, opened 2023-03-01
no news is good news

sastels:
# Summary | Résumé

Set the malware scanning timeout warning to not breaching (so no timeouts means all is good)


andrewleith (approved):
Looks good, thanks!

PR #678, opened 2023-03-06
add regex for static assets

sastels:
# Summary | Résumé

In #676 we forgot to add paths to static assets.

Tested the regex by manually adding to the staging WAF.

Merge and ensure the [staging documentation site](https://documentation.staging.notification.cdssandbox.xyz/) works (When the page loads you should not get any 204s).


jimleroyer (approved):
Thank you @sastels !

PR #681, opened 2023-03-07
fix: lambda admin PR review env security group

patheard:
# Summary
Update the lambda admin PR review environment function to have its own security group.  This will make it easier to apply fine grained access control to the lambda and Redis cluster.

Also updates the lambda's security group to give it access to the VPC private endpoints and receive requests from the internet on port 443.

# Related
- cds-snc/notification-planning#650
- cds-snc/notification-planning#1017
- cds-snc/platform-core-services#285


jimleroyer (approved):
LGTM

PR #684, opened 2023-03-10
Relax scan-files timeout warning alert to 3 events within 5 minutes

jimleroyer:
# Summary | Résumé

We are getting too many warning noise around scan-file at the moment. The files scan can take more than 11 minutes, at least to what we receive on our end. Until we find the root cause and resolve, we will relax this alarm. 

I checked historical data to see what we had so far. We can get 2 scanfiles-timeout events within 5 minutes, but not 3. Hence why I established this threshold within these changes to have no noise about this unless current state degrades further.


ben851 (approved):
Looks good! Thank you for doing this, I was getting concerned with the erroneous alerts causing alert fatigue

smcmurtry (approved):

PR #689, opened 2023-03-22
New relic force secret name

ben851:
# Summary | Résumé

Trying a better solution to get the new relic license key to be read properly by the api lambda. This will force the lambda function to read the license key from the secret defined in the aws secret manager, "NEW_RELIC_LICENSE_KEY".  Trying this in case the lambda function was attempting to read from a different secret value.

> Deploy to staging and see if new relic continues to report properly

### DIFF
--- a/aws/lambda-api/lambda.tf
+++ b/aws/lambda-api/lambda.tf
@@ -40,7 +40,6 @@ resource "aws_lambda_function" "api" {
       NEW_RELIC_EXTENSION_LOGS_ENABLED      = true
       NEW_RELIC_LAMBDA_EXTENSION_ENABLED    = true
       NEW_RELIC_LICENSE_KEY                 = data.aws_secretsmanager_secret_version.new-relic-license-key.secret_string
-
       FF_CLOUDWATCH_METRICS_ENABLED         = var.ff_cloudwatch_metrics_enabled
     }
   }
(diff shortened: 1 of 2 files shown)


aws/lambda-api/lambda.tf
```diff
@@ -39,6 +39,7 @@ resource "aws_lambda_function" "api" {
       NEW_RELIC_DISTRIBUTED_TRACING_ENABLED = var.new_relic_distribution_tracing_enabled
       NEW_RELIC_EXTENSION_LOGS_ENABLED      = true
       NEW_RELIC_LAMBDA_EXTENSION_ENABLED    = true
+      NEW_RELIC_LICENSE_KEY_SECRET          = "NEW_RELIC_LICENSE_KEY"
```
sastels, line 42: Won't this just set it to the string? Or is this the point, to prevent the parameter store variables to put the value in?

ben851: It's a bit confusing but this is telling new relic to contact aws secret manager and get the secret from the secret key of "NEW_RELIC_LICENSE_KEY". [As per this doc.](https://docs.newrelic.com/docs/serverless-function-monitoring/aws-lambda-monitoring/enable-lambda-monitoring/account-linking/)

sastels: ok makes sense now!

sastels (approved):

PR #694, opened 2023-03-23
Revert "Force lambda function to read new relic license from env variable"

ben851:
Reverts cds-snc/notification-terraform#688

Removing changes to see if New relic will read the license key correctly.


sastels:
Don't know why we have these iam changes, it's nothing in this PR 🤔

PR #697, opened 2023-03-30
Increasing the monthly spend limit for SNS on staging

ben851:
# Summary | Résumé

We have been very close to the spending quota for SNS text messaging in staging. AWS has increased our quota (case: 12368522451). This PR sets new spend limits to $100/month in staging.


sastels (approved):
:shipit:

PR #698, opened 2023-03-30
Revert "Increasing the monthly spend limit for SNS on staging"

ben851:
Doesn't look like our AWS quota was increased - I'll revert this for now and check w/ AWS Support

Reverts cds-snc/notification-terraform#697


andrewleith (approved):
✅

PR #702, opened 2023-04-04
Chore/tag all the things

sastels:
# Summary | Résumé

As described in https://github.com/cds-snc/opa_checks/blob/main/aws_terraform/tags.rego, tag our resouces with (at minimum) 
```
CostCentre = notification-canada-ca-staging (or prod)
Terraform = true
```

Look at the results of the plan


sastels:
going to wait until default tagging works, helpfully won't be long

PR #703, opened 2023-04-04
Edit IAM permissions for new relic user

jzbahrai:
# Summary | Résumé

Task: [#49](https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/49)

Add permissions for new relic user with regard to the API lambda.

### DIFF
--- a/aws/lambda-api/iam.tf
+++ b/aws/lambda-api/iam.tf
@@ -84,6 +84,15 @@ data "aws_iam_policy_document" "api_policies" {
     effect = "Allow"
     actions = [
       "secretsmanager:GetSecretValue",
+    ]
+    resources = [
+      aws_secretsmanager_secret_version.new-relic-license-key.arn
+    ]
+  }
+
+  statement {
+    effect = "Allow"
+    actions = [
       "cloudwatch:GetMetricStatistics",
       "cloudwatch:ListMetrics",
       "cloudwatch:GetMetricData",
@@ -96,9 +105,7 @@ data "aws_iam_policy_document" "api_policies" {
       "lambda:ListTags",
       "lambda:ListEventSourceMappings"
     ]
-    resources = [
-      aws_secretsmanager_secret_version.new-relic-license-key.arn
-    ]
+    resources = ["*"]
   }
 }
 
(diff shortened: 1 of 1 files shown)


ben851:
This looks good, although I'm concerned as to how this is currently working in both prod and staging without these permissions assigned to the policy... Unless there is a global policy that was done via clickops at some point. Lets try it in staging to see how it goes.

aws/lambda-api/iam.tf
```diff
@@ -84,6 +84,17 @@ data "aws_iam_policy_document" "api_policies" {
...
+      "cloudwatch:ListMetrics",
+      "cloudwatch:GetMetricData",
+      "config:BatchGetResourceConfig",
+      "config:ListDiscoveredResources",
+      "tag:GetResources",
+      "lambda:GetAccountSettings",
+      "lambda:ListFunctions",
+      "lambda:ListAliases",
+      "lambda:ListTags",
+      "lambda:ListEventSourceMappings"
```
jimleroyer, line 97: Is this the right place to add these? @ben851 indicated that the actions should be applied on `*` resources. Would the the new relic key resource need to do these actions? I might be misunderstanding something, iam is always some black magic to me.

ben851: I think this is good, it's where I had them for mine. I am going off what was here:
https://docs.newrelic.com/docs/serverless-function-monitoring/aws-lambda-monitoring/enable-lambda-monitoring/account-linking/

New Relic's doc's are a bit confusing in that this is talking about linking aws accounts but underneath it's talking about using the lambda integration...

ben851: Ok I blew it you're 100% right Jimmy. I missed that this was assigned to the secret arn only. I will request the change

ben851 (changes requested):
As per my conversation with @jimleroyer, these permissions are not scoped to the secret manager ARN - they need to be scoped to all resources (*).  This can be done by simply creating a new block in data "aws_iam_policy_document" "api_policies"   adding these new values and setting the resources to *

ben851 (approved):
Looks good!

PR #705, opened 2023-04-04
Adding a warning for K8s node failure

ben851:
# Summary | Résumé

Creating a cloudwatch anomaly alert which will trigger if node failure occurs.  Currently set to warning while we verify that it works as expected.

### DIFF
--- a/aws/eks/cloudwatch_alarms.tf
+++ b/aws/eks/cloudwatch_alarms.tf
@@ -357,15 +357,14 @@ resource "aws_cloudwatch_metric_alarm" "logs-3-scanfiles-timeout-5-minutes-warni
   alarm_actions       = [var.sns_alert_warning_arn]
 }
 
-
 resource "aws_cloudwatch_metric_alarm" "kubernetes-failed-nodes" {
-  alarm_name                = "kubernetes-failed-nodes"
-  comparison_operator       = "GreaterThanUpperThreshold"
-  evaluation_periods        = 1
-  threshold_metric_id       = "ad1"
-  alarm_description         = "Kubernetes failed node anomalies"
+  alarm_name          = "kubernetes-failed-nodes"
+  comparison_operator = "GreaterThanUpperThreshold"
+  evaluation_periods  = 1
+  threshold_metric_id = "ad1"
+  alarm_description   = "Kubernetes failed node anomalies"
   #Setting to warn until we verify that it is working as expected
-  alarm_actions             = [var.sns_alert_warning_arn]
+  alarm_actions = [var.sns_alert_warning_arn]
 
   metric_query {
     id          = "ad1"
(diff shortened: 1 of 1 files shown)


aws/eks/cloudwatch_alarms.tf
```diff
@@ -356,3 +356,36 @@ resource "aws_cloudwatch_metric_alarm" "logs-3-scanfiles-timeout-5-minutes-warni
...
+    return_data = "true"
+  }
+
+  metric_query {
+    id          = "m1"
+    return_data = "true"
+    metric {
+      metric_name = "cluster_failed_node_count"
+      namespace   = "ContainerInsights"
+      period      = 300
```
jimleroyer, line 382: Would we have a faster alarm raised if we lower the period here?

aws/eks/cloudwatch_alarms.tf
```diff
@@ -356,3 +356,36 @@ resource "aws_cloudwatch_metric_alarm" "logs-3-scanfiles-timeout-5-minutes-warni
...
+  threshold_metric_id = "ad1"
+  alarm_description   = "Kubernetes failed node anomalies"
+  #Setting to warn until we verify that it is working as expected
+  alarm_actions = [var.sns_alert_warning_arn]
+
+  metric_query {
+    id          = "ad1"
+    expression  = "ANOMALY_DETECTION_BAND(m1, 2)"
+    label       = "cluster_failed_node_count (expected)"
+    return_data = "true"
```
jimleroyer, line 373: Why are we passing the pick to an anomaly detection expression from `m1` to `ad1` to the alarm threshold? Why not do `m1` to the alarm threshold directly?

ben851: Disclaimer, I'm not 100% familiar with cloudwatch queries yet.

The way I determined how to code the query was by doing it via click-ops in my scratch environment and then exporting the query JSON, which I then transcoded to TF format. So what you see here is the way AWS built the query.

If I understand it correctly what is happening is that the m1m query is running every 5 minutes and returning a result. The AD1 query is then querying the results of M1 using the Anomaly detection algorithm to look for these anomalies across the time series of m1 results. We can't simply use the anomaly detection query as the primary query as there would be no time series in that case. 

If you see something I'm missing please let me know :)

jimleroyer (changes requested):
Requesting change to get rid of anomaly detection and have the metric rely on the direct metric, as we do not require a trend detection (but mere existence of the failure).

aws/eks/cloudwatch_alarms.tf
```diff
@@ -356,3 +356,28 @@ resource "aws_cloudwatch_metric_alarm" "logs-3-scanfiles-timeout-5-minutes-warni
...
 }
+
+resource "aws_cloudwatch_metric_alarm" "kubernetes-failed-nodes" {
+  alarm_name          = "kubernetes-failed-nodes"
+  comparison_operator = "GreaterThanThreshold"
+  evaluation_periods  = 1
+  alarm_description   = "Kubernetes failed node anomalies"
+  #Setting to warn until we verify that it is working as expected
+  alarm_actions      = [var.sns_alert_warning_arn]
+  treat_missing_data = "notBreaching"
```
jimleroyer, line 367: Yes we are learning from our previous noise. 🤝 😅

jimleroyer (approved):
LGTM! :shipit:

PR #706, opened 2023-04-05
global replace of CostCenter with CostCentre

sastels:
# Summary | Résumé

Bring our tags closer to the CDS standard.

Here we just do a global search and replace of CostCenter with CostCentre.


sastels:
Only `common` and `performance-test` have destroys, the rest just have changes.

PR #712, opened 2023-04-18
Standardize treat_missing_data on alarms after review

jimleroyer:
# Summary | Résumé

We are reviewing how alarms should treat missing data: should these breach, raise a warning or ignore lacks of data points? It depends on the source of the alarms. We adjusted based on historic observations.

| Alarm type        | Resolution    | Reasoning                                                                                                      |
|-------------------|---------------|----------------------------------------------------------------------------------------------------------------|
| SQS               | missing       | SQS is reliable and rarely miss data points. Surfacing as warnings.                                            |
| EKS Pod Resources | missing       | Missing data points for the pods could indicate an underlying issue with the pods. Surfacing as warnings.      |
| Lambda SES        | not breaching | These were observed to miss data points once in a while. No noise.                                             |
| Lambda API logs   | not breaching | These were observed to miss data points once in a while. No noise.                                             |
| Redis             | missing       | These were observed to miss data points once in a while. Surfacing as warnings but we'll keep an eye on these. |
| SES               | missing       | SES is reliable and rarely miss data points. Surfacing as warnings.                                            |

The changes also add actions for insufficient data points. We realized as we were reviewing the alarms that we only receive the status changes from `INSUFFICIENT DATA -> OK`, but never `OK -> INSUFFICIENT DATA`. We will be testing that to see how it behaves and might adjust further.

* Make sure the transition state `OK -> INSUFFICIENT DATA` works.
* Have these changes stay in staging environment for 1 day and observe the noise level. The goal is to reduce and have more of the one that matter.


aws/eks/cloudwatch_alarms.tf
```diff
@@ -428,9 +444,8 @@ resource "aws_cloudwatch_metric_alarm" "celery-sms-replicas-unavailable" {
       stat        = "Average"
       unit        = "Count"
       dimensions = {
-        ClusterName = aws_eks_cluster.notification-canada-ca-eks-cluster.name
```
jimleroyer, line 431: Do we need to remove this?

aws/common/cloudwatch_alarms.tf
```diff
@@ -468,9 +492,32 @@ resource "aws_cloudwatch_metric_alarm" "sqs-send-throttled-sms-tasks-receive-rat
...
+  threshold          = 60 * 1.1
+  treat_missing_data = "missing"
+  alarm_actions      = [aws_sns_topic.notification-canada-ca-alert-warning.arn]
+  ok_actions         = [aws_sns_topic.notification-canada-ca-alert-ok.arn]
+  dimensions = {
+    QueueName = "${var.celery_queue_prefix}${var.sqs_throttled_sms_queue_name}"
+  }
+}
+
+resource "aws_cloudwatch_metric_alarm" "sqs-send-throttled-sms-tasks-receive-rate-critical" {
```
jimleroyer, line 504: This is a new alarm, needs 👀

jzbahrai (approved):
you are the best and deserve all the best things in the world. thank you for doing this 🙇🏽

aws/common/cloudwatch_alarms.tf
jzbahrai: this makes sense to me to start with.

PR #713, opened 2023-04-18
change some alerts from max to avg

sastels:
# Summary | Résumé

Change the priority queue delay alarm metrics from `Maximum` to `Average`. Hopefully quiets down to false alarms we've had.

This is hopefully a temporary fix until we rework our queues.


ben851 (approved):

jimleroyer (changes requested):
I think we should only change to average for the priority queue. The approximate oldest message + average for other queues will make it much more tolerant to longer time for notifications and break our SLOs. Let's talk offline.

jimleroyer (approved):
LGTM

PR #714, opened 2023-04-18
Remove unit=count from promotheus dimensions

sastels:
# Summary | Résumé

Removing the count unit in promotheus dimensions.

I compared sources of non-working TF metric to a working one in staging and this was the one change that made the difference. Hoping this TF changes make the metric work. 🤞

* Make an API pod crash in staging env and observe if the alarm triggers.


jimleroyer (approved):
Hey it looks good to me!!

PR #716, opened 2023-04-20
upgrade redis node type

sastels:
# Summary | Résumé

[Embiggen Redis Cluster](https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/46
)

Upgrade redis node type from t3.micro to t3.small (so 500 MB ->  1.37 GB memory)


sastels:
terraform plan says it'll update in place. Should run a perf test while doing the upgrade and verify that no notifications are lost.

sastels:
Not needed right now - looks like we'll stay under our current memory limits for the foreseeable future.

PR #717, opened 2023-04-20
Adding config file for aws-nuke in scratch environment

ben851:
# Summary | Résumé

This is the configuration file for running [aws-nuke](https://github.com/rebuy-de/aws-nuke) in upcoming scratch environments. This has no impact on staging or production environments, and is currently explicitly set to my scratch environment. We can modify this config file in the future to include additional scratch accounts. This is more of a utility PR that will be used when terragrunt fails to properly delete. 

aws-nuke --config scratch-nuke.yaml --profile ben-scratch

### DIFF
--- a/env/scratch/aws-nuke/scratch-nuke.yaml
+++ b/env/scratch/aws-nuke/scratch-nuke.yaml
@@ -1,0 +1,197 @@
+# Regions to remove resources from
+regions:
+- "global"
+- "us-east-1"
+- "us-west-2"
+- "ca-central-1"
+
+# Only delete resources from accounts that do not match LandingZone filters
+
+accounts:
+  "419291849580": 
+      filters:
+        IAMPolicy:
+        - type: glob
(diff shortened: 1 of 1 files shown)


env/scratch/aws-nuke/scratch-nuke.yaml
```diff
@@ -0,0 +1,197 @@
...
+    - CloudFrontOriginAccessIdentity
+    - CloudFrontDistribution
+    - IAMVirtualMFADevice
+    - AppStreamImage
+
+
+
+
+# Accounts that will not have resources removed
+account-blocklist:
```
sastels, line 161: Do we need this, since you explicitly say what account to nuke? or is this a prudent fail-safe?

ben851: I stole those from the SRE team's aws-nuke config, I assume you are correct in that it's a failsafe approach.. Better safe than sorry!

PR #721, opened 2023-04-20
Refactoring Terraform For Scratch Deployment

ben851:
# Summary | Résumé

Lots of changes here.
* [Had to update AWS Provider and modify s3 bucket creation due to this breaking change by AWS.](https://github.com/hashicorp/terraform-provider-aws/issues/28353) This needs to be verified in staging.
* Removed reference to static cloud based sensor satellite bucket, and instead created this bucket via TF. We will need to import this in staging and production, and further changes may be necessary to configure the bucket to forward to the CBS.
* Made the athena workgroup name a variable - it looks like the name "primary" is created by default in the AWS account. The default name is "primary" but for scratch environment, this can be overridden (and has to be since the primary already exists).
* Set the monthly spending limit for scratch to $1

* Against scratch env, terragrunt apply -var-file ../scratch.tfvars
* Terragrunt plan against staging/prod and import as necessary.


ben851:
The failing CI tests are because the aws provider  in staging has to be upgraded to 4.0, which requires a terragrunt init --upgrade command. Don't want to do this yet.

aws/common/s3.tf
```diff
@@ -478,3 +535,81 @@ resource "aws_s3_bucket" "athena_bucket_logs" {
...
+  depends_on = [
+    aws_s3_bucket_public_access_block.athena_bucket_logs,
+    aws_s3_bucket_ownership_controls.athena_bucket_logs,
+  ]
+
+  bucket = aws_s3_bucket.athena_bucket_logs.id
+  acl    = "log-delivery-write"
+}
+
+resource "aws_s3_bucket" "cbs_sensor_bucket" {
```
CalvinRodo, line 565: Have you seen our Terraform modules? 

https://github.com/cds-snc/terraform-modules

We have modules for S3 Buckets that might be useful. 

When v5 of the AWS Provider comes out and this type of bucket is finally deprecated I plan on fixing the s3 bucket module so it automatically refactors to the new way of building them. 

It also configures them in a consistent that meets some of our organizational defined best practices. 

It's not 100% perfect so open to any issues that can help us secure and make our orgs infra better.

CalvinRodo (commented):
Just some non-blocking comments pointing you to our Terraform Modules.

aws/common/s3.tf
ben851: Looks good Calvin. I'd probably still proceed with this PR for now, but once we're back into a state where things actually build I'll do some investigation in to what modules we can leverage. Happy to be a contributor to make them as useful and secure as possible 👍

jimleroyer (approved):
LGTM

aws/common/s3.tf
patheard, line 565: Right now the CBS bucket is being managed here for all accounts:
https://github.com/cds-snc/cloud-based-sensor

If you want to bring it into this repo you'll need to do a bit of IAM setup and add replication rules so it can send objects to the log archive account's bucket:
https://github.com/cds-snc/cloud-based-sensor/tree/main/terragrunt/aws/satellite_bucket

Happy to chat about this if you want more details.

ben851: Thanks Pat - based on this info I think I will refactor this.

ben851:
I'm going to close this PR for now - I'm going to look at refactoring the code for the CBS Satellite buckets.

PR #724, opened 2023-04-24
Imported SQS queue to production, bumping version

ben851:
# Summary | Résumé

* Importing existing celery internal tasks sqs queue to terraform
* Bumping infrastructure version

Terragrunt plan on common should show no changes.


jimleroyer:
@ben851 I'll check it out after that release.

PR #725, opened 2023-04-24
Removing EBS encryption reference from state in staging

ben851:
# Summary | Résumé

The EBS encryption enabled account wide was being set within the EKS cluster section of terraform. We are unable to destroy this resource as per CDS landing zone policies. I raised an issue with the CDS SRE team to have EBS encryption enabled in the landing zone code, and they just completed this. We can now safely remove the reference from our own code as it is configured in theirs. 

terragrunt state rm aws_ebs_encryption_by_default.notification-canada-ca


jimleroyer (approved):
Looks like it does as advertised.

PR #732, opened 2023-04-27
Common S3 Fix

ben851:
# Summary | Résumé

Reworking CBS satellite logs bucket to be conditionally created, migrating existing log_buckets to CDS S3 logs module

* terragrunt plan/apply in staging
* verify log buckets are still working


aws/common/s3.tf
```diff
@@ -24,7 +25,7 @@ resource "aws_s3_bucket" "csv_bucket" {
 
   #tfsec:ignore:AWS077 - Versioning is not enabled
   logging {
-    target_bucket = aws_s3_bucket.csv_bucket_logs.bucket
+    target_bucket = regex("^[^.]*", module.csv_bucket_logs.s3_bucket_domain_name)
```
patheard, line 28: You should be able to use the [module's `s3_bucket_id` output](https://github.com/cds-snc/terraform-modules/blob/24e4de7bbb94139504fed050bb26f5ee55e105d5/S3_log_bucket/outputs.tf#L1-L4) for this:
```suggestion
    target_bucket = module.csv_bucket_logs.s3_bucket_id
```

aws/common/s3.tf
```diff
@@ -91,7 +80,7 @@ resource "aws_s3_bucket" "bulk_send" {
 
   #tfsec:ignore:AWS077 - Versioning is not enabled
   logging {
-    target_bucket = aws_s3_bucket.bulk_send_logs.bucket
+    target_bucket = regex("^[^.]*", module.bulk_send_logs.s3_bucket_domain_name)
```
patheard, line 83: Same comment about the `s3_bucket_id` module output.

aws/common/s3.tf
```diff
@@ -241,7 +210,7 @@ resource "aws_s3_bucket" "document_bucket" {
 
   #tfsec:ignore:AWS077 - Versioning is not enabled
   logging {
-    target_bucket = aws_s3_bucket.document_bucket_logs.bucket
+    target_bucket = regex("^[^.]*", module.document_download_logs.s3_bucket_domain_name)
```
patheard, line 213: Same comment about the `s3_bucket_id` module output.

aws/common/variables.tf
```diff
@@ -211,3 +211,21 @@ variable "alarm_critical_expired_email_created_threshold" {
   description = "Critical alarm threshold for expired email batches created"
   type        = number
 }
+
+variable "force_destroy_s3" {
```
patheard, line 215: Nice, this'll be handy based on the number of times `terragrunt destroy` has failed for me with S3 buckets.

aws/common/s3.tf
ben851: 100% correct - forgot that aws treats id and bucket name as the same thing. Fixed.

sastels:
Code and plan lgtm.

aws/common/vpc.tf
```diff
@@ -203,6 +203,9 @@ resource "aws_default_network_acl" "notification-canada-ca" {
 }
 
 resource "aws_flow_log" "cloud-based-sensor" {
+  depends_on = [
+    module.cbs_logs_bucket
+  ]
```
patheard, line 208: Non-blocking comment, but I'm curious how this'll behave for an environment where `var.create_cbs_bucket = false`.  I've never tried a `depends_on` block with a conditional resource.

patheard (approved):

aws/common/vpc.tf
ben851: Yeah I checked that because I also was not 100% sure - it works. I believe the reasoning behind it working is as follows: when it sees the depends on block, it waits for the evaluation of that dependency. When false, that dependency evaluation returns a "count = 0, and therefore resource deployment is complete" (aka dependency fulfilled)

env/scratch/common/terragrunt.hcl
```diff
@@ -7,8 +7,8 @@ include {
 }
 
 inputs = {
-  sns_monthly_spend_limit                                            = 50
-  sns_monthly_spend_limit_us_west_2                                  = 30
+  sns_monthly_spend_limit                                            = 1
+  sns_monthly_spend_limit_us_west_2                                  = 1
```
jimleroyer, line 11: Good move! 👍

jimleroyer (approved):
LGTM!

env/scratch/common/terragrunt.hcl
ben851: Can't deploy to a new account without it 😄  Manual intervention is required to up the limit.

ben851:
Verified with Steve that he's good to go with these changes.

PR #733, opened 2023-04-27
Tune sqs-priority-queue-delay-critical

whabanks:
# Summary | Résumé
When a notification in the priority queue fails, we retry the notification after a 5 minutes. This results in the `sqs-priority-queue-delay-critical` alarm triggering a false positive.

The alarm will now only trigger if the 60 second threshold is breached for five, one minute periods within a 5 minute evaluation period.


jimleroyer (approved):
LGTM :shipit:

PR #743, opened 2023-05-09
Acm prod migration

ben851:
# Summary | Résumé

* Production certificates migrated to new Terraform locations
* Had to update IAM policy for staging route53 to allow production to talk to it. This seems to be a terraform limitation that I will investigate
* Set route53 records to only create in staging for now (waiting on prod access)

Terragrunt plan in production should show no changes


aws/cloudfront/dns.tf
```diff
@@ -1,4 +1,6 @@
 resource "aws_route53_record" "assets-notification-CNAME" {
+  count = var.env != "production" ? 1 : 0
```
sastels, line 2: Why not check against "staging" ? Is it so they'll be created in scratch?

ben851: Yep exactly... we need them to be created in everything not production for now. Once we have access to production dns zone, this count will go away

sastels (approved):
🚀

PR #747, opened 2023-05-11
Fixing ACM validation by iterating over SANs and creating route53

ben851:
# Summary | Résumé

The route53 records in the acm validation config were only creating the first entry. If there are SANs, they validation records were not being created, and thus cert validation was failing. This change adds all records.

terragrunt apply in staging.

### DIFF
--- a/aws/cloudfront/acm.tf
+++ b/aws/cloudfront/acm.tf
@@ -15,19 +15,30 @@ resource "aws_acm_certificate" "assets-notification-canada-ca" {
 }
 
 resource "aws_route53_record" "assets-notification-canada-ca" {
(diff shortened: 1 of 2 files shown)


sastels:
So I guess this LGTM but what do I know 😬  I see you've just added Pat which I was going to suggest might be a good idea!

ben851:
> So I guess this LGTM but what do I know 😬 I see you've just added Pat which I was going to suggest might be a good idea!

The good news is that if this doesn't work, it's a very smooth roll back, which is the next step if this doesn't go

patheard (approved):

PR #750, opened 2023-05-11
Custom domain mx record

ben851:
# Summary | Résumé

After migrating to the new notify managed staging DNS, we get a warning notice from AWS saying that the custom-domain MX record was missing from notify, and was required for SES. I am adding it back here.

[ ] Terragrunt Apply in staging
[ ] MX Record shows up in notify sandbox DNS


sastels (approved):
LGTM will we be able to see if this makes AWS happy?

PR #753, opened 2023-05-16
Changing staging alt domain to be under notification.cdssandbox.xyz

ben851:
# Summary | Résumé

Modifying staging alt domain to be underneath the notification.cdssandbox.xyz domain. This is required because the ACM validation needs to be able to create route53 records on the alt domain, which I have not mapped in staging (it is shared with other apps). This will not have any impact on prod, since we will not be migrating the DNS.

* Terragrunt plan/apply should work
* ACM Validation should work in staging

### DIFF
--- a/env/staging/env_vars.hcl
+++ b/env/staging/env_vars.hcl
@@ -1,7 +1,7 @@
 inputs = {
   account_id = "239043911459"
   domain     = "staging.notification.cdssandbox.xyz"
-  alt_domain = "staging.notification.alpha.cdssandbox.xyz"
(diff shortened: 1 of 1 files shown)


sastels (approved):
seems simple enough!
Will this make staging.alpha.notification.cdssandbox.xyz exist? or do we have to do something else?

ben851:
> seems simple enough! Will this make staging.alpha.notification.cdssandbox.xyz exist? or do we have to do something else?

Yes! Because we now own notification.cdssandbox.xyz, terraform can (and will) automatically add/update the DNS records.

PR #754, opened 2023-05-16
Adding DNS records for alt-dns in staging

ben851:
# Summary | Résumé

Adding the alternate domain DNS record entries into staging dns.

Terragrunt apply in staging works
Validate that new DNS entries are resolvable

### DIFF
--- a/aws/eks/dns.tf
+++ b/aws/eks/dns.tf
@@ -27,6 +27,36 @@ resource "aws_route53_record" "notificatio-root-WC" {
 
 }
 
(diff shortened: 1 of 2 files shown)


sastels (commented):
Should we be a bit concerned with the
```
Warning: Argument is deprecated
```
ie is there a different argument we should be using?

ben851:
> Should we be a bit concerned with the
> 
> ```
> Warning: Argument is deprecated
> ```
> 
> ie is there a different argument we should be using?

sastels (approved):
:shipit:

PR #755, opened 2023-05-17
SES DKIM validation dns records

ben851:
# Summary | Résumé

Adding back the SES DKIM validation records after not migrating them from the SRE notification.cdssandbox.xyz route53. We got an alert from AWS saying these were required continually.

This required a number of complex changes/reorganization:
* Because the custom domain and tryapply dkim zones are indexed by count, I had to add a nested for loop to pull the correct validation records out of the indexed entries and store them in a flat object
* The flat object itself had to then be rebuilt as a  key value map to be able to iterate over them.
* Due to the above, the validation objects had to be removed from the DNS folder. Terraform could not create net-new environments as it lacks the logic to understand that it needs to create an object (dkim domain) before building the validation token variables above in locals. 
* Added a new folder ses_validation_dns_entries that will get run after dns is run. Added the appropriate steps to staging github actions (production will follow when DNS access for prod is granted)

---

* Validate the records are created in the new aws route53 zone (staging)
* AWS should notify us that the records have been restored


sastels (approved):
ok, looks like the plan's saying it'll do the right thing, let's see what apply thinks!

PR #756, opened 2023-05-17
fix: increase s3_scan_object function timeout

patheard:
# Summary
Update to the latest `s3_scan_object` module version which increases the transport lambda's timeout to 120s.  This is to fix the cases where the transport lambda must wait for a new Scan Files lambda function cold start.

This will fix the timeout errors currently occurring and will hopefully also stop the `scanfiles-timeout` errors being seen:

image

# Related
- cds-snc/platform-core-services#297
- cds-snc/terraform-modules#251


jimleroyer (approved):
LGTM

ben851 (approved):
That's one heck of a version bump!

PR #760, opened 2023-05-18
Reseting the scan file timeout alarm to initial value after applied fix

jimleroyer:
# Summary | Résumé

* Reseting the scan file timeout alarm to initial value after applied fix

Let it work in staging for a few days.


sastels (approved):
🚀

PR #764, opened 2023-05-24
Lambda merge

ben851:
# Summary | Résumé

This is (hopefully) the final PR that has the entire notify environment build/destroy from scratch. It's a large PR, but there is lots of repetition following a set pattern (for each of the lambdas as an example). 

- [ ] Terragrunt plan in staging shows minimal to no changes (Some resources may need to be moved/migrated)
- [ ] Terragrunt plan in production shows minimal to no changes  (Some resources may need to be moved/migrated)
- [ ] Terragrtunt apply in the following order builds an environment from scratch
    - common
    - ecr
    - ses_receiving_emails
    - dns
    - ses_validation_dns_entries
    - cloudfront
    - eks
    - elasticache
    - rds
    - lambda-api
    - lambda-admin-pr
    - performance-test
    - heartbeat
    - database-tools
    - lambda-google-cidr
    - ses_to_sqs_email_callbacks
    - sns_to_sqs_sms_callbacks
- [ ] Terragrtunt destroy in reverse of the above order completes successfully


ben851:
Looks like this will have to be a progressive migration

aws/common/sqs.tf
```diff
@@ -28,4 +28,18 @@ resource "aws_sqs_queue" "notify_internal_tasks_queue" {
   sqs_managed_sse_enabled = false
   # This queue was created outside of terraform and has this value set to 310 in staging and production.
   visibility_timeout_seconds = 310
+}
+
+resource "aws_sqs_queue" "eks_notification_canada_cadelivery_receipts" {
+  name             = "eks-notification-canada-cadelivery-receipts"
```
jimleroyer, line 34: There is a case where we can use prefix to create custom queues that gets outside of the normal naming convention, to run locally when developing. Raising it in case that might be a problem as these queues wouldn't get managed in Terraform.
image

aws/ecr/images.tf
```diff
@@ -0,0 +1,205 @@
+#Notify Admin Build and Push
```
jimleroyer, line 1: Good comment, could we add some explanations why we are doing this way? 🙏

aws/lambda-api/cloudwatch_logs.tf
```diff
@@ -12,6 +12,18 @@ resource "aws_cloudwatch_log_group" "api_gateway_log_group" {
   }
 }
 
+resource "aws_cloudwatch_log_group" "api_lambda_log_group" {
+  name              = "/aws/lambda/${aws_lambda_function.api.function_name}"
+  retention_in_days = 90
```
jimleroyer, line 17: 90 days only?

aws/lambda-api/secrets_manager.tf
```diff
@@ -1,5 +1,12 @@
+resource "random_string" "new_relic_postfix" {
+  count = var.env == "production" || var.env == "staging" ? 0 : 1
+
+  length  = 8
+  special = false
+}
+
 resource "aws_secretsmanager_secret" "new-relic-license-key" {
-  name        = "NEW_RELIC_LICENSE_KEY"
+  name        = var.env == "production" || var.env == "staging" ? "NEW_RELIC_LICENSE_KEY" : "NEW_RELIC_LICENSE_KEY_${random_string.new_relic_postfix[0].result}"
```
jimleroyer, line 9: Hmm if it's not staging or production, should we blank this instead? as a random key wouldn't work with NewRelic?

aws/performance-test/secrets.tf
```diff
@@ -1,5 +1,29 @@
...
+
+resource "random_string" "perf_test_auth_header_postfix" {
+  count   = var.env == "production" || var.env == "staging" ? 0 : 1
+  length  = 8
+  special = false
+}
+
 resource "aws_secretsmanager_secret" "perf_test_phone_number" {
-  name = "perf_test_phone_number"
+  name = var.env == "production" || var.env == "staging" ? "perf_test_phone_number" : "perf_test_phone_number_${random_string.perf_test_phone_number_postfix[0].result}"
```
jimleroyer, line 26: 1. We don't have `perf_test_phone_number` defined for production, only staging. 
2. I don't think we'd want to use a randomly generated number, because we risk of trying to send notifications to this number and pay for these. The test number used in staging comes from a [specially flagged phone number](https://documentation.notification.canada.ca/en/testing.html) that won't get sent (so we do not pay).

aws/performance-test/secrets.tf
```diff
@@ -8,7 +32,7 @@ resource "aws_secretsmanager_secret_version" "perf_test_phone_number" {
 }
 
 resource "aws_secretsmanager_secret" "perf_test_email" {
-  name = "perf_test_email"
+  name = var.env == "production" || var.env == "staging" ? "perf_test_email" : "perf_test_email${random_string.perf_test_email_postfix[0].result}"
```
jimleroyer, line 35: This would only get defined in staging too... 

Is the random generation gets appended to `perf_test_email`?

aws/performance-test/secrets.tf
```diff
@@ -17,7 +41,7 @@ resource "aws_secretsmanager_secret_version" "perf_test_email" {
 }
 
 resource "aws_secretsmanager_secret" "perf_test_domain" {
-  name = "perf_test_domain"
+  name = var.env == "production" || var.env == "staging" ? "perf_test_domain" : "perf_test_domain${random_string.perf_test_auth_header_postfix[0].result}"
```
jimleroyer, line 44: Same would apply here as well (only staging needs it).

aws/performance-test/secrets.tf
```diff
@@ -26,7 +50,7 @@ resource "aws_secretsmanager_secret_version" "perf_test_domain" {
 }
 
 resource "aws_secretsmanager_secret" "perf_test_auth_header" {
-  name = "perf_test_auth_header"
+  name = var.env == "production" || var.env == "staging" ? "perf_test_auth_header" : "perf_test_auth_header${random_string.perf_test_auth_header_postfix[0].result}"
```
jimleroyer, line 53: Same would apply here as well (only staging needs it).

scripts/clonerepo.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
jimleroyer, line 1: Where is this file called from?

aws/common/sqs.tf
ben851: We can include a new flag called "create_local_queues" or something and have that queue be created by TF at scratch deployment. Would that resolve this?

aws/ecr/images.tf
ben851: Sure I can add in a bit of text explaining that the ECR and docker image need to exist before the lambda function can be built.

aws/lambda-api/cloudwatch_logs.tf
ben851: I'm not sure I wrote 90. They're set to never expire in staging and prod. I will adjust this.

aws/performance-test/secrets.tf
ben851: 1. Ok I'll remove the prod part of the check
2. The random postfix here is not for the value of the number - that would still be set how we currently do. This is just for the name (key) part of the secret. I had to add postfixes because AWS secrets have an account wide scope, meaning that if we ever wanted to set up multiple environments in the same account, TF would fail at this step because the resource already exists.

aws/performance-test/secrets.tf
ben851: Will do the same as above :)

aws/performance-test/secrets.tf
ben851: Will do the same as above :)

scripts/clonerepo.sh
ben851: This was part of another option I was exploring for the docker images. I'll get rid of it.

aws/lambda-api/secrets_manager.tf
ben851: The flag to enable/disable new relic broke the lambda function, so I've omitted this for now. I'll create a card to further investigate the relationship between the app and new relic at deploy time.

jimleroyer: Alrighty

jimleroyer (approved):

PR #771, opened 2023-05-30
Prod sync2

ben851:
# Summary | Résumé

This PR will sync prod and staging in prep for new infra version release.


jzbahrai (approved):

aws/ecr/ecr.tf
```diff
@@ -49,6 +50,7 @@ resource "aws_ecr_repository" "performance-test" {
   # The :latest tag is used in Staging
   #tfsec:ignore:AWS078
 
+  count                = var.env == "production" ? 0 : 1
```
jzbahrai, line 53: i dont know what this does :/

ben851: This says if the environment is production, don't build a performance test ECR since that's only for staging :)

PR #784, opened 2023-06-04
fix: upgrade to SQS queue scan files

patheard:
# Summary
Upgrade the `s3_scan_object` module to the version that uses an SQS queue to deliver S3 object notifications to the Scan Files API.

In performance testing this has resulted in a substantially higher throughput for S3 object create events as they can be batched in groups of 10 and do not get constrained by transport lambda throttling errors.

# Related
- cds-snc/platform-core-services#297


ben851 (approved):
LGTM!

PR #787, opened 2023-06-05
Bumping k8s version to 1.24

ben851:
# Summary | Résumé

Bumping EKS K8s up to 1.24 from 1.23. Will need to run a soak test against staging once this is merged before going to production.

* Soak test against staging while terraform is performing the upgrade
* Smoke test against staging once the upgrade is complete to verify system integrity


jimleroyer (approved):
LGTM

PR #789, opened 2023-06-07
add new GAC sending domains

sastels:
# Summary | Résumé

re https://cds-snc.freshdesk.com/a/tickets/15122

Add new sending domains @international.gc.ca and @core-ocre.gc.ca

After this merges, need to give various records to the user as described in [this ticket](https://cds-snc.freshdesk.com/a/tickets/7468).


jimleroyer (approved):
Let's try this! :shipit:

PR #792, opened 2023-06-08
release 1.3.5

sastels:
# Summary | Résumé

Basically want to add the new sending domains to prod.

[Reset scanfiles alarm to more restrictive alert #791
](https://github.com/cds-snc/notification-terraform/pull/791)
[add new GAC sending domains #789
](https://github.com/cds-snc/notification-terraform/pull/789)

PR #793, opened 2023-06-08
Revert "release 1.3.5"

sastels:
Reverts cds-snc/notification-terraform#792

This will remove the new sending domains that could trigger celery errors as we saw earlier. 

We have a fix, but smoke tests are failing consistently (timing out) on staging so I'd rather not push more code from staging to production.


sastels:
Well, the plan here doesn't want to remove the new sending domains 😞

sastels:
well, this PR didn't really revert the changes and there haven't been more alerts about it so I guess we can keep them in 🤞

PR #794, opened 2023-06-09
Removing manual references to sql alchemy URIs 

ben851:
# Summary | Résumé

Instead of having to manually specify the sql alchemy URIs and storing them in TFVars files, we can now retrieve them programmatically via terraform. This will simplify new environment deployment, reduce the number of secrets required in tfvars, and reduce the chance of configuration errors.

Tested in scratch account. 

Staging and production will create new versions of the sql alchemy secrets in parameter store and recycle the lambdas to pick up the new value even though the value will remain the same (TF thinks it's changed since the way it's generated is different)

### DIFF
--- a/aws/database-tools/parameterstore.tf
+++ b/aws/database-tools/parameterstore.tf
@@ -38,7 +38,7 @@ resource "aws_ssm_parameter" "notify_o11y_google_oauth_client_secret" {
 resource "aws_ssm_parameter" "sqlalchemy_database_reader_uri" {
   name  = "sqlalchemy_database_reader_uri"
   type  = "SecureString"
-  value = "postgresql://postgres:${var.rds_cluster_password}@read-only-endpoint.${var.database_proxy_endpoint}"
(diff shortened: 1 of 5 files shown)


aws/database-tools/parameterstore.tf
```diff
@@ -38,7 +38,7 @@ resource "aws_ssm_parameter" "notify_o11y_google_oauth_client_secret" {
 resource "aws_ssm_parameter" "sqlalchemy_database_reader_uri" {
   name  = "sqlalchemy_database_reader_uri"
   type  = "SecureString"
-  value = var.sqlalchemy_database_reader_uri
```
sastels, line 41: These should be using the `app_db_user` and `app_db_user_password` rather than the postgres user.

sastels (changes requested):

aws/database-tools/parameterstore.tf
ben851: Yup fixed.

PR #797, opened 2023-06-13
release 1.3.7

sastels:
# Summary | Résumé

Construct the sqlalchemy uris from other data already available.
- #794
- #795
- #796


ben851 (approved):
I'm going to create a PR to fix that RDS log change thing but this can go for now!

PR #799, opened 2023-06-13
Rds param fix

ben851:
# Summary | Résumé

Adding in the apply method of pending-reboot for rds log retention since AWS continually sets it to this value. The default is immediate but this particular parameter must be configured as pending-reboot.

After deploying this fix, terraform should not longer show a change to the rds parameter group when no change was done.


sastels (approved):
🚀

PR #802, opened 2023-06-14
capitalizing E on emails

ben851:
# Summary | Résumé

Need to capitalize E on "Emails" dashboard so that we don't have to delete it in prod to recreate. In staging it was lower case.

Verify that dashboard is named "Emails" in staging after apply (existing dashboard will be destroyed)
Verify that no changes are required on PR to prod


whabanks (approved):
It's beautiful - LGTM!

PR #798, opened 2023-06-13
Dashboards

ben851:
# Summary | Résumé

I added all the dashboards in production to the TF code so we can replicate them in staging. This is in support of the alarm cleanup.

# Test instructions | Instructions pour tester la modification

TF Apply works, dashboards work in staging.

PR #804, opened 2023-06-14
Release 1.3.10

whabanks:
# Summary | Résumé
- #798
- #799
- #802


jimleroyer (changes requested):
Please let's add the PRs in the description that upgrade this version, as a good practice to track down what went down into a release. 🙏

PR #808, opened 2023-06-16
Converted to base URL instead of individual URL definitions

ben851:
# Summary | Résumé

Instead of having to define individual URLs for all of the different endpoints, I have created a base_domain (and alt_base_domain) variable that can be set once to the root domain. From there, TF will automatically create the appropriate URLs.

Tested in scratch.
Test in staging, TF Plan should show minimal changes (since the domains are the same).  There will be a recreate on api gateway due to a weird trigger based on the file contents, and not the actual values that are sent.


ben851:
> CHANGE	NAME
> recreate	`aws_api_gateway_deployment.api`
> update	`aws_api_gateway_stage.api`

This is that weird re-creation that is triggered based off of whether or not the checksum of the TF file has changed. Since I have modified the file (switching variables), it wants to recreate. I've verified it works in scratch. Eventually I'd like to investigate why this is necessary to begin with.

> CHANGE	NAME
> update	`module.heartbeat.aws_lambda_function.this`

I verified that this change is changing from double quotes to single quotes. Single quotes is better in this case, since it doesn't require escape characters. The underlying URLs are the same.

sastels (approved):
LGTM

PR #815, opened 2023-06-21
Less noise from the ses br alarms

jimleroyer:
# Summary | Résumé

* Treat the SES bounce rate alarms as non breaching on missing data.
* Converted all period field values into integers.
* Increased service's bounce rate warning alarm from 1 minute to 3 minutes.

* Regularly checks in #notification-staging-ops Slack channel if the ses br alarms still go off.
* Checks on the warning alarms when it triggers and how it behaved in the AWS dashboard.

### DIFF
--- a/aws/eks/cloudwatch_alarms.tf
+++ b/aws/eks/cloudwatch_alarms.tf
@@ -11,7 +11,7 @@ resource "aws_cloudwatch_metric_alarm" "load-balancer-1-500-error-1-minute-warni
   evaluation_periods  = "1"
   metric_name         = "HTTPCode_ELB_500_Count"
   namespace           = "AWS/ApplicationELB"
-  period              = "60"
+  period              = 60
   statistic           = "Sum"
   threshold           = 1
   alarm_actions       = [var.sns_alert_warning_arn]
@@ -28,7 +28,7 @@ resource "aws_cloudwatch_metric_alarm" "load-balancer-10-500-error-5-minutes-cri
   evaluation_periods  = "1"
   metric_name         = "HTTPCode_ELB_500_Count"
   namespace           = "AWS/ApplicationELB"
-  period              = "300"
+  period              = 300
   statistic           = "Sum"
   threshold           = 10
   alarm_actions       = [var.sns_alert_critical_arn]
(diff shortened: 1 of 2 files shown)


aws/eks/cloudwatch_alarms.tf
```diff
@@ -11,7 +11,7 @@ resource "aws_cloudwatch_metric_alarm" "load-balancer-1-500-error-1-minute-warni
   evaluation_periods  = "1"
   metric_name         = "HTTPCode_ELB_500_Count"
   namespace           = "AWS/ApplicationELB"
-  period              = "60"
+  period              = 60
```
andrewleith, line 14: Are all these just syntax changes, then? Not related to BR right?

aws/eks/cloudwatch_alarms.tf
```diff
@@ -380,21 +380,21 @@ resource "aws_cloudwatch_metric_alarm" "logs-1-bounce-rate-1-minute-critical" {
...
   statistic           = "Sum"
   threshold           = 1
   treat_missing_data  = "notBreaching"
   alarm_actions       = [var.sns_alert_warning_arn]
 }
 
 resource "aws_cloudwatch_metric_alarm" "logs-1-bounce-rate-1-minute-warning" {
-  alarm_name          = "logs-1-warning-bounce-rate-1-minute-warning"
-  alarm_description   = "One service exceeding 5% bounce rate in 1 minute"
+  alarm_name          = "logs-1-warning-bounce-rate-3-minutes-warning"
```
andrewleith, line 391: These values make sense to me.  One thing I was wondering the other day was why we don't end up with TONS of alarms when a service goes past the threshold and keeps sending?  In this case, we continually write the same log statements so I was surprised (but happy) that it didn't happen.  Any ideas why?

For example, one service in particular had 8034 critical bounce rate logs written, but I don't think the alarms went haywire.

andrewleith (approved):
LGTM - a question on some seemingly unrelated syntax changes, and a question on how the alarms work when there are thousands of logs written.

whabanks (approved):

aws/eks/cloudwatch_alarms.tf
jimleroyer: The alarm sends a notification using a SNS topic (which will trigger a Slack notification) when it switches from one state to the other, hence `OK -> In Alarm -> Ok`, totaling 2 notifications for the 2 switches (or arrows). Repeated errors while in the `In Alarm` state will keep the alarm in the same state. 

For repeated notification on a periodic basis, a [quick googling](https://aws.amazon.com/blogs/mt/how-to-enable-amazon-cloudwatch-alarms-to-send-repeated-notifications/) does not show this to be out of the box, but possible using EventBridge with a rule that would trigger a periodic process when an alarm is triggered, hence basically feeding it from a SNS event and EventBridge refeeding the some other SNS events.

PR #822, opened 2023-06-29
fix: remove Salesforce errors from API error metric

patheard:
# Summary
Update the Lambda API error metric to no longer trigger for Salesforce errors, which are identified by the `SF_ERR` token.

This will stop both alarms from triggering if a Salesforce error is detected.


sastels (approved):
I guess this excludes lines with `SF_ERR` ?

Is there an easy way to trigger a Salesforce error on staging or prod to test this?

patheard:
Yup that's exactly it, and you can test this by clicking into the log group and then going to the `Metric filters` tab:
image

From here you can click into the metric filter you want to test and provide sample log output to see if it works as expected.  I'm planning to do that once this applies.

PR #823, opened 2023-06-29
Scan object to reference s3 bucket by ID

jimleroyer:
# Summary | Résumé

* Made the scan object S3 object name shorter to make room for scratch accounts customizations.
* Reference the scan object s3 bucket by ID.

* Plan and apply should work.


ben851:
Hmm interesting - the bucket name in staging is different from the other environments...

ben851 (approved):
After speaking with Jimmy, Steve M, and Pat, we determined that the recreation of the s3 bucket for scan_files shouldn't be an issue. We will test this in staging once merged.

PR #824, opened 2023-06-30
Quicksight

sastels:
# Summary | Résumé

Add AWS Quicksight

**Note:**
- requires a [new RDS user to first be created](https://docs.google.com/document/d/1BkAkz45CVQRGQSNwN018beTebrbTUNpyTpGEaO7Clyc/edit#heading=h.mnku3slnkhf2) (done on staging)
- only plans / applies the Quicksight module in staging


aws/quicksight/data_sources.tf
```diff
@@ -0,0 +1,23 @@
+
+resource "aws_quicksight_data_source" "default" {
```
ben851, line 2: Can we change the resource name from "default" to something more meaningful? Even just "Notify"

aws/quicksight/data_sources.tf
```diff
@@ -0,0 +1,23 @@
...
+      password = var.quicksight_db_user_password
+    }
+  }
+  vpc_connection_properties {
+    vpc_connection_arn = aws_quicksight_vpc_connection.rds.arn
+  }
+  parameters {
+    rds {
+      database    = var.database_name
+      instance_id = "test"
```
ben851, line 19: What is an instance_id in this case? If it's an arbitrary value, can we sync it to var.env?

aws/quicksight/vpc.tf
```diff
@@ -0,0 +1,41 @@
...
+          ]
+          Resource = ["*"]
+        }
+      ]
+    })
+  }
+}
+
+resource "aws_quicksight_vpc_connection" "rds" {
+  vpc_connection_id  = "example-connection-id"
```
ben851, line 36: this will need to be the vpc_id from the "common" outputs.

aws/quicksight/vpc.tf
```diff
@@ -0,0 +1,41 @@
...
+  name               = "Example Connection"
```
ben851, line 37: Can we change this to something like "Notify VPC"

ben851 (changes requested):
Very good first crack, just some small refinements to make

aws/quicksight/data_sources.tf
sastels: hrm, when I was playing around before it was the `aws_db_instance` identifier. We have a cluster identifier now maybe? I'll try that...

ben851 (approved):
LGTM!

PR #827, opened 2023-07-11
Empty log alarm for lambda api

ben851:
# Summary | Résumé

This will send an alert to the Notify warning slack channel when no Lambda API logs are detected over a 10 minute period. The metric seems to only be updated once every 5 minutes, so a 10 minute window ensures we have at least some data points.

Deploy to staging, validate that alarm does not trigger during regular use. Disable lambda api to trigger alarm.


sastels:
We probably want to add alarms for the k8s apps as well (though probably elsewhere in the repo). Were you going to do that in a different PR?

sastels (approved):
LGTM (I didn't know the SRE folks had this nifty module!). Just one question.

jimleroyer (approved):
That's great, thanks for the addition.

ben851:
> We probably want to add alarms for the k8s apps as well (though probably elsewhere in the repo). Were you going to do that in a different PR?

Unfortunately the module doesn't work for K8s - we can't get granular enough. I've created a task to figure out how to do this in for k8s as well

PR #828, opened 2023-07-12
fix: Terraform module version references

patheard:
# Summary
Update the Terraform module version references so they are in the [correct format](https://developer.hashicorp.com/terraform/language/modules/sources#modules-in-package-sub-directories): 

> If the source address has arguments, such as the ref argument supported for the version control sources, the sub-directory portion must be before those arguments.

This will allow Renovate dependency PRs to update the module versions without stripping the `//sub-directory` path.

# Note
There should be no TF plan changes.

### DIFF
--- a/aws/common/file_scanning.tf
+++ b/aws/common/file_scanning.tf
@@ -1,5 +1,5 @@
 module "s3_scan_objects" {
-  source = "github.com/cds-snc/terraform-modules?ref=v6.0.1//S3_scan_object"
+  source = "github.com/cds-snc/terraform-modules//S3_scan_object?ref=v6.0.1"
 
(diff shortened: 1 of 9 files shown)


jimleroyer (approved):
LGTM!

ben851 (approved):
LGTM!

sastels:
looks fine to me - checked before and after for any new copy-paste errors.

How did this work before? 🤔

patheard:
> looks fine to me - checked before and after for any new copy-paste errors.
> 
> How did this work before? 🤔

I honestly have no idea - according to the docs, it shouldn't have, but it's how we've also always done it.

PR #829, opened 2023-07-12
Cloudwatch alarms for evicted pods in k8s

ben851:
# Summary | Résumé

Adding alarms that will trigger when any of the notify apps show evicted pods.

**NOTE: A manual import of the prometheus cloudwatch log group will be required for staging and prod.**

Tested in scratch.

Unfortunately there doesn't seem to be an easy way to test this exactly. I verified that the queries return the correct data in staging on June 16th. I also tested the inverse query (kube_pod_status_reason = 0) to validate that the alarm trips.


sastels (approved):
LGTM! Is there a way to manually evict to test?

ben851:
> LGTM! Is there a way to manually evict to test?

From what I've been able to see, unfortunately not in the way we need :( When a pod gets evicted by K8s, it stays showing in the namespace as status evicted, but when you manually tell k8s to evict a pod, it just deletes it...

PR #830, opened 2023-07-12
Prod dns migration prep

ben851:
# Summary | Résumé

This contains a copy/paste of the cds DNS repository entry for notification.canada.ca where I then removed the values that are now dynamically generated as well as the individual document download and documentation entries, which are now caught by a wildcard A record to go to the K8s load balancer.

There are still several hard coded values in notification.canada.ca that need to be migrated eventually but are currently manual in other environments as well.

I have renamed the staging provider in AWS to dns since we will use the same provider entry with a different account id for production. 

I have also updated the DNS records for staging and scratch api endpoints to be monitored by the DNS zone for health as this is the configuration for production.

**Note that creating these resources in production will not make them live. We still require SSC to do the DNS switch**

- Terragrunt apply in production works
- A manual comparison of the existing production route 53 zone with the newly created resources will need to be done in order to validate that the values are as expeceted.

### DIFF
--- a/env/terragrunt.hcl
+++ b/env/terragrunt.hcl
@@ -7,7 +7,7 @@ inputs = {
   domain             = "${local.vars.inputs.domain}"
   alt_domain         = "${local.vars.inputs.alt_domain}"
   env                = "${local.vars.inputs.env}"
-  staging_account_id = "${local.vars.inputs.staging_account_id}"
+  dns_account_id     = "${local.vars.inputs.dns_account_id}"
   region             = "ca-central-1"
   # See https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html#access-logging-bucket-permissions
   elb_account_ids = {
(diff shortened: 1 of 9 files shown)


env/terragrunt.hcl
```diff
@@ -47,16 +47,25 @@ provider "aws" {
...
 
+provider "aws" {
+  alias  = "dns"
+  region = "ca-central-1"
+  assume_role {
+    role_arn = "arn:aws:iam::${local.vars.inputs.dns_account_id}:role/${local.vars.inputs.env}_dns_manager_role"
+  }
+}
+
 provider "aws" {
```
ben851, line 58: Note that the old staging provider has to stay in place until the new terragrunt apply is done. Once complete, we can remove this.

sastels:
@patheard set a lot of the old stuff up - would be good to get his 👀 on this

jimleroyer (approved):

PR #832, opened 2023-07-13
task(blazer+slack): pass a slack webhook secret into the blazer ecs task

andrewleith:
# Summary | Résumé

This PR passes an existing webhook secret over to blazer for use it its configuration.  Once functionality is tested and working, we can plumb in the correct webhook (which still needs to be created).


ben851 (approved):
LGTM!

PR #837, opened 2023-07-18
DNS IAM Fix

ben851:
# Summary | Résumé

Since the DNS provider needs the IAM controls for DNS management to already be in place, they must be moved out of the DNS directory. I have moved them to common for now, but it may be beneficial to create a "prerequisites" folder that has all the account pre-reqs.

TF Apply in staging should recreate the DNS IAM
TF Apply in prod should create new IAM for prod DNS
TF Plan for DNS Prod Migration will then work.

### DIFF
--- a/aws/common/iam-dns-prereqs.tf
+++ b/aws/common/iam-dns-prereqs.tf
@@ -87,7 +87,9 @@ EOF
 }
 
 resource "aws_iam_role_policy" "production_dns_manager_policy" {
-  count = var.env == "production" ? 1 : 0
+  # Putting this on both prod and staging, since currently prod is creating some junk staging records.
+  # Once Prod DNS migration is completed, the staging portion can be removed
+  count = var.env == "production" || var.env == "staging" ? 1 : 0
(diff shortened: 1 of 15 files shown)


env/production/prereqs/terragrunt.hcl
```diff
@@ -0,0 +1,8 @@
+terraform {
```
jimleroyer, line 1: oh I am not sure I get this file, we are referencing the DNS sources and then include parent folders? What is the actual requirement in there, just the defined variables in the end?

ben851: Yeah every folder needs a terragrunt.hcl for terragrunt to understand what's happening... This one is really simple because there are no dependencies... but it still needs the find in parent folder to get the common variables and providers from the root terragrunt.hcl.

jimleroyer (approved):
LGTM (but I'm kinda clueless)

ben851:
> LGTM (but I'm kinda clueless)

Happy to get on a google meet and whiteboard it out if you'd like!

PR #839, opened 2023-07-19
Dns provider fix

ben851:
# Summary | Résumé

Removing cross account dns roles in terragrunt.hcl when the environment is staging or production. They are not needed since the DNS zone that is being modified is in the same account. This also fixes the chicken/egg scenario where terragrunt fails to apply because it requires account roles that don't yet exist.

Terragrunt apply in staging/production work as expected

### DIFF
--- a/env/terragrunt.hcl
+++ b/env/terragrunt.hcl
@@ -1,5 +1,6 @@
 locals {
   vars = read_terragrunt_config("../env_vars.hcl")
+  dns_role           = local.vars.inputs.env == "production" || local.vars.inputs.env == "staging" ? "" : "assume_role { role_arn = \"arn:aws:iam::${local.vars.inputs.dns_account_id}:role/${local.vars.inputs.env}_dns_manager_role\" }"
 }
 
 inputs = {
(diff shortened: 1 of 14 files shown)


env/terragrunt.hcl
```diff
@@ -1,5 +1,7 @@
 locals {
   vars = read_terragrunt_config("../env_vars.hcl")
+  # dns_role is very fragile - if not set exactly as below, terraform fmt will fail in github actions.
+  dns_role           = local.vars.inputs.env == "production" || local.vars.inputs.env == "staging" ? "" : "\n  assume_role {\n    role_arn = \"arn:aws:iam::${local.vars.inputs.dns_account_id}:role/${local.vars.inputs.env}_dns_manager_role\"\n  }"
```
jimleroyer, line 4: 🧙

env/terragrunt.hcl
```diff
@@ -1,5 +1,7 @@
 locals {
   vars = read_terragrunt_config("../env_vars.hcl")
+  # dns_role is very fragile - if not set exactly as below, terraform fmt will fail in github actions.
```
jimleroyer, line 3: Can we also add a comment on why we need this?

env/terragrunt.hcl
```diff
@@ -105,6 +105,11 @@ variable "cbs_satellite_bucket_name" {
   description = "Name of the Cloud Based Sensor S3 satellite bucket"
   type        = string
 }
+variable "cross_account_dns" {
+  description = "Set to true if the DNS zone is not in the same account as the target aws account"
+  type        = bool
+  default     = true
+}
```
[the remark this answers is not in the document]
ben851: Whoops this is test code. I will remove.

jimleroyer (approved):
This is quite satisfying as the amount of changes are way less than previous attempts and that is more elegant as well. Good job!

PR #841, opened 2023-07-20
Revert "scan object to reference s3 bucket by ID (#823)"

jimleroyer:
# Summary | Résumé

This raises the document attachment to fail in staging as we need extra changes, and the CI pipeline is already blocked by other changes. So we are reverting that one to reduce the bottleneck and reducing risk. We will get back to it later.

Send a notification with a file attachment in staging once deployed and check if the scan files functionality is working as expected through the CW logs.


ben851 (approved):
As noted, I would like line 4 of file_scanning.tf to reference the created TF object but since that was not in the original code, we can revert as is for now and then make a new PR to fix that.

PR #844, opened 2023-07-20
Route 53 variable

ben851:
# Summary | Résumé

The variable for the route 53 zone was named inconsistently across the project. I have consolidated it all to use the singular route_53_zone_arn variable name.

TF plan/apply works as expected with correct zones being populated for staging vs prod.


jimleroyer (approved):
:shipit:

PR #852, opened 2023-07-28
Creating procedure, and scripts for EKS node upgrades

ben851:
# Summary | Résumé

Creating the logic, scripts, and documentation for upgrading EKS nodes without downtime.  We can upgrade or downgrade node sizes at will.

- Logic and procedure tested in scratch
- Another team member should QA this in staging.


jimleroyer:
That looks good to me. I'll let @sastels throw his eyes on this as well before approving. This is handy to have, especially for the documentation, so that the 🚌 OKR is getting done. 👍

scripts/cordonAndDrain.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+NODE_GROUP=$1
+NODES=$(kubectl get nodes -l eks.amazonaws.com/nodegroup=$NODE_GROUP | awk 'NR>1{print $1}')
+echo "Begin cordoning nodes"
+while IFS= read -r node; do
+    echo "Cordoning node $node"
+    kubectl cordon $node
```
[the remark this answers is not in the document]
ben851: It's amazing what google and stack overflow can achieve! 😃

docs/nodeUpgrade.md
```diff
@@ -0,0 +1,68 @@
...
+- Commit these changes, and create a PR
+- Validate in the PR that a new node group will be created and that none will be destroyed
+- Merge the PR
+- Verify that the new nodes are created (kubectl get nodes)
+
+#### Migrate Workload to Secondary Node Group
+
+- From a local machine with administrator account access to AWS, run the cordonAndDrain.sh script in the scripts directory, passing in the name of the **primary** node group  
+    Example:
+    ``` ./cordonAndDrain.sh notification-canada-ca-scratch-eks-primary-node-group ```
```
sastels, line 40: curious: how does it know the name of where we're migrating **to**? just assumes there's one other node group and uses that?

ben851: Yeah, when we spin up the new node group, it auto-registers with EKS. So all we do here is still it to stop using the old ones, and, since the new nodes are already registered, it just moves to those ones instead.

sastels (approved):
LGTM too! as discussed IRL we'll make sure we do a soak test on staging when we test this the first time.

PR #853, opened 2023-07-31
set force_destroy=true for depecated bulk_send buckets

sastels:
# Summary | Résumé

We want to delete these buckets but first have to allow `force_destroy` on them.


ben851 (approved):
LGTM

PR #861, opened 2023-08-02
upgrade primary staging nodes to r5a.large

sastels:
# Summary | Résumé

Goal: switch k8s nodes from m6i.large to r5a.large. (see [docs](https://github.com/cds-snc/notification-terraform/blob/main/docs/nodeUpgrade.md))

this step: upgrade primary to r5a.large

PR #862, opened 2023-08-02
delete secondary node group

sastels:
# Summary | Résumé

Goal: switch k8s nodes from m6i.large to r5a.large. (see [docs](https://github.com/cds-snc/notification-terraform/blob/main/docs/nodeUpgrade.md))

previous step: #861 
this step: delete secondary node group


ben851 (approved):
LGTM!

PR #867, opened 2023-08-03
Karpenter

ben851:
# Summary | Résumé

This is the PR to put in place the IAM configuration for Karpenter in the notificaiton-manifests. 
Karpenter is an open source auto-scaler for kubernetes.

This is more of a for your review PR instead of a "lets merge it"

For your review

### DIFF
--- a/aws/eks/iam.tf
+++ b/aws/eks/iam.tf
@@ -2,6 +2,8 @@
 # AWS EKS IAM cluster roles
 ###
 
+
+
 resource "aws_iam_role" "eks-cluster-role" {
   name = "eks-cluster-role"
 
@@ -13,6 +15,13 @@ resource "aws_iam_role" "eks-cluster-role" {
       "Effect": "Allow",
       "Principal": {
         "Service": "eks.amazonaws.com"
+      },
+      "Action": "sts:AssumeRole"
+    },
+    {
+      "Effect": "Allow",
+      "Principal": {
+        "Service": "ec2.amazonaws.com"
       },
       "Action": "sts:AssumeRole"
     }
@@ -158,6 +167,7 @@ resource "aws_iam_role_policy_attachment" "notification-fargate-worker-policy" {
 }
 
 
+
 ###
 # AWS EKS Service account
 ###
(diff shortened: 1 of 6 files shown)


aws/eks/iam.tf
```diff
@@ -198,4 +208,88 @@ resource "aws_iam_role" "notification-service-account-role" {
...
+  depends_on = [aws_eks_cluster.notification-canada-ca-eks-cluster]
+}
+
+data "aws_iam_policy" "ssm_managed_instance" {
+  arn = "arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore"
+}
+
+
+module "iam_assumable_role_karpenter" {
+  source                        = "terraform-aws-modules/iam/aws//modules/iam-assumable-role-with-oidc"
```
jimleroyer, line 229: So this module comes directly from AWS? or from a 3rd party?

aws/eks/iam.tf
```diff
@@ -198,4 +208,88 @@ resource "aws_iam_role" "notification-service-account-role" {
...
+
+resource "aws_iam_role_policy_attachment" "karpenter_ssm_policy" {
+  role       = aws_iam_role.eks-worker-role.name
+  policy_arn = data.aws_iam_policy.ssm_managed_instance.arn
+}
+resource "aws_iam_role_policy_attachment" "karpenter-cluster-worker-node" {
+  policy_arn = "arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy"
+  role       = module.iam_assumable_role_karpenter.iam_role_name
+}
+resource "aws_iam_role_policy_attachment" "karpenter-cluster-cni" {
```
jimleroyer, line 250: What does CNI means?

jimleroyer: Cloud Network Interface is my best guess so far.

https://github.com/aws/amazon-vpc-cni-k8s

aws/eks/iam.tf
```diff
@@ -198,4 +208,88 @@ resource "aws_iam_role" "notification-service-account-role" {
...
+          "ec2:DescribeInstances",
+          "ec2:DescribeInstanceTypes",
+          "ec2:DescribeInstanceTypeOfferings",
+          "ec2:DescribeAvailabilityZones",
+          "ec2:DeleteLaunchTemplate",
+          "ec2:CreateTags",
+          "ec2:CreateLaunchTemplate",
+          "ec2:CreateFleet",
+          "ec2:TerminateInstances",
+          "ec2:DescribeSpotPriceHistory",
```
jimleroyer, line 287: So can Karpenter select spot instances or others based on budget parameters?

jimleroyer (commented):
That looks great. 👍

aws/eks/iam.tf
ben851: Not quite on budget parameters - if choosing spot instances, it goes for the best bang-for-buck that will meet the resource requirements while also ensuring that the spots aren't about to be terminated. If it's on-demand pricing, then it just picks lowest cost. [Details can be found here](https://karpenter.sh/docs/faq/) under "How does Karpenter dynamically select instance types?"

What we can do is "provision ONLY these instance types" and then have a pre-defined list of instance types that are within our budget.

aws/eks/iam.tf
ben851: It's actually container network interface!
https://github.com/containernetworking/cni

aws/eks/iam.tf
jimleroyer: Ah that is pretty cool. 👍

aws/.checkov.yml
```diff
@@ -19,4 +19,5 @@ skip-check:
   - CKV_AWS_31  # TODO: Elasticache investigate in-transit data encryption and auth token  
   - CKV_AWS_91  # TODO: Load balancer enable access logging
   - CKV2_AWS_8  # TODO: RDS investigate if Query Logging can be enabled
-  - CKV2_AWS_27 # TODO: RDS investigate how to setup AWS Backup plan
\ No newline at end of file
+  - CKV2_AWS_27 # TODO: RDS investigate how to setup AWS Backup plan
+  - CKV_TF_1 # You cannot reference commit hashes on terraform registry.
```
jimleroyer, line 23: Why would Chekhov recommends against that? 🤔

aws/eks/iam.tf
```diff
@@ -261,6 +261,11 @@ resource "aws_iam_role_policy_attachment" "karpenter-cluster-managed-instance-co
 }
 
 resource "aws_iam_role_policy" "karpenter_controller" {
+  #checkov:skip=CKV_AWS_290:The Karpenter IAM requires blanket access
+  #checkov:skip=CKV_AWS_286:The Karpenter IAM requires privilge escalation
```
jimleroyer, line 265:
```suggestion
  #checkov:skip=CKV_AWS_286:The Karpenter IAM requires privilege escalation
```

aws/eks/iam.tf
jimleroyer: That's great we have a specific version then! Thanks!

jimleroyer (approved):
LGTM (there is a typo fix suggested though)

aws/.checkov.yml
ben851: Best practice is to reference commit SHA for versioning since tags are mutable... But Terraform Registry doesn't support this (yet). The idea being that someone could create a malicious version of the module and somehow inject it into the github repo (or TF registry)... Of course the account holders of the repo would have to be compromised for that to happen.

PR #869, opened 2023-08-08
Add a script to cordon and drain one node and tweak node group script

sastels:
# Summary | Résumé

Script to cordon and drain one node.
Also tweaked the existing script to add a pause between drains


ben851 (approved):
Looks good!

PR #874, opened 2023-08-11
add a quicksight vpc connection for connection to rds

sastels:
# Summary | Résumé

add back the quicksight vpc connection (maybe!)


ben851 (approved):
Verified that it deploys successfully on scratch

PR #877, opened 2023-08-14
add alarms for new send-sms queues

sastels:
# Summary | Résumé

In
https://github.com/cds-snc/notification-api/pull/1958
we move sending sms to its own queues, send-sms-high, send-sms-medium, and send-sms-low.

Here we essentially copy the alarms for the queues the corresponding tasks were previously in, priority-tasks, send-sms-tasks, and bulk-tasks.


aws/common/cloudwatch_alarms.tf
```diff
@@ -336,6 +336,115 @@ resource "aws_cloudwatch_metric_alarm" "sqs-sms-stuck-in-queue-critical" {
...
+  insufficient_data_actions = [aws_sns_topic.notification-canada-ca-alert-warning.arn]
+  ok_actions                = [aws_sns_topic.notification-canada-ca-alert-ok.arn]
+  dimensions = {
+    QueueName = "${var.celery_queue_prefix}${var.sqs_send_sms_medium_queue_name}"
+  }
+}
+
+resource "aws_cloudwatch_metric_alarm" "sqs-send-sms-low-queue-warning" {
+  alarm_name          = "sqs-send-sms-low-queue-warning"
+  alarm_description   = "ApproximateAgeOfOldestMessage in send-sms-low queue is >= 60 minutes"
```
jimleroyer, line 414: That seems fair to model the warning/critical alarms of SMS low priority the same as emails, but  admittedly, we didn't really consider the timing for these. Let's keep it at these values for now but let's also update the values in the [SLO 1.5 document](https://docs.google.com/spreadsheets/d/1cJjCzqC9JOAq005jWg9OjmeYoQvLkPaAFkCaIZTcaWA/edit#gid=0) please. 🙏

jimleroyer (approved):
Thank you!

aws/common/cloudwatch_alarms.tf
sastels: hrm, yeah, I just copied the old queues. Since we have separate SMS queues I think it makes sense to change the thresholds to the SLO values rather than change the SLO, at least for warnings. How about if I change the warnings to:
high: 10 seconds for 3 minutes 
med: 10 minutes for 5 minutes
high: 10 minutes for 5 minutes

PR #879, opened 2023-08-14
Fix QuickSight datasource: use a different identifier

sastels:
# Summary | Résumé

Was (I think?) using the identifier for the wrong thing...


ben851 (approved):
This feels right.

Coding is based on emotions right?

PR #886, opened 2023-08-17
set prod node group storage to 80G

sastels:
# Summary | Résumé

Upgrade prod node storage from 29G to 80G.

Tested on staging.


ben851 (approved):
LGTM!

PR #883, opened 2023-08-16
Secondary target group

ben851:
# Summary | Résumé

This is the first PR to support fixing the EKS load balancer services without downtime as documented here:
https://github.com/cds-snc/notification-manifests/pull/1947

# Test instructions | Instructions pour tester la modification

- Terraform apply should add new target groups
- Verify that all hosted URLs on EKS still work as expected:
   - https://staging.notification.cdssandbox.xyz
   - https://api-k8s.staging.notification.cdssandbox.xyz
   - https://documentation.staging.notification.cdssandbox.xyz
   - https://api.document.staging.notification.cdssandbox.xyz/_status

PR #887, opened 2023-08-17
EKS Load Balancer Fix - Stage 1

ben851:
# Summary | Résumé

This PR will create secondary target groups on the EKS ALB but they will not be used. This is the first step of the split up of https://github.com/cds-snc/notification-terraform/pull/883 Which failed.

Apply to staging, validate that staging still works w/ a soak test.


jimleroyer:
This is only to create these atm?

PR #905, opened 2023-09-06
Setting Cloudwatch as Feature Flag

ben851:
# Summary | Résumé

This will add an option to disable most cloudwatch log groups, alarms, and dashboards. This is to create some cost savings for scratch accounts where rigorous cloudwatch logs are not required. In the event that we would need to test cloudwatch functionality in scratch, we can simply set the feature flag cloudwatch_enabled to true. 

The default value is true, so there should be no changes to staging or production

- Deployed and tested in scratch
- Staging plan should show at maximum that it must recreate lambda functions (it may not do this if it is smart enough)
- Smoke test staging
- Terraform may state that a resource has moved since we are now using counts and keys for resources.


aws/common/cloudwatch_log.tf
```diff
@@ -51,10 +56,11 @@ resource "aws_cloudwatch_log_group" "route53_resolver_query_log" {
 # AWS CloudWatch Logs Metrics
 ###
 resource "aws_cloudwatch_log_metric_filter" "sns-sms-blocked-as-spam" {
-  name = "sns-sms-blocked-as-spam"
+  count = var.cloudwatch_enabled ? 1 : 0
+  name  = "sns-sms-blocked-as-spam"
   # See https://docs.amazonaws.cn/en_us/sns/latest/dg/sms_stats_cloudwatch.html#sms_stats_delivery_fail_reasons
   pattern        = "{ $.delivery.providerResponse = \"Blocked as spam by phone carrier\" }"
-  log_group_name = aws_cloudwatch_log_group.sns_deliveries_failures.name
+  log_group_name = aws_cloudwatch_log_group.sns_deliveries_failures[0].name
```
jimleroyer, line 63: Hmm was that not working before?

ben851: It was working, but because I'm now setting the creation of the log group conditional to the feature flag, I have to reference it as an array now (which yes, is an array of one - thanks Terraform)

ben851:
> delete	`module.lambda_no_log_detection.aws_cloudwatch_metric_alarm.empty_log_group_metric_alarm[&quot;/aws/lambda/api-lambda&quot;]`
> add	`module.lambda_no_log_detection[0].aws_cloudwatch_metric_alarm.empty_log_group_metric_alarm[&quot;/aws/lambda/api-lambda&quot;]`

This looks like a symptom of a resource in a nested module not cleanly transferring into the array. It's not a big deal to delete and recreate it, especially since it's just an alarm, and not a core piece of functionality.

jimleroyer (approved):
LGTM!

PR #910, opened 2023-09-07
Checkov

ben851:
# Summary | Résumé

This will switch the (not working) TF Sec scanning check to use Checkov instead. Checkov will also automatically publish the results to the [GitHub code scanning tracker](https://github.com/cds-snc/notification-terraform/security/code-scanning?query=pr%3A910+is%3Aopen). 

I have configured Checkov to run in permissive mode so that the pipelines don't fail when there are failing tests. At the moment there are a large number of failing tests that we will have to triage and either fix, or add to the ignore list should they be deemed false alarms.


.github/workflows/terragrunt_plan_staging.yml
```diff
@@ -301,4 +301,4 @@ jobs:
           comment-delete: "true"
           comment-title: "Staging: sns_to_sqs_sms_callbacks"
           github-token: "${{ secrets.GITHUB_TOKEN }}"
-          terragrunt: "true"
+          terragrunt: "true"
```
sastels, line 304: Github is happiest when there's a blank line at the end 😂

.gitignore
```diff
@@ -6,6 +6,7 @@
 *.tfstate.*
 *.tfvars
 *.yaml
+!.checkov.yaml
```
sastels, line 9: The file has a ".yml" extension so it would probably be included anyway? Or were you intending to name with a ".yaml"

sastels (approved):
LGTM! Could add those blank lines to get GitHub happy and change the ".yml" suffix (or change the .gitignore) if you wise

.gitignore
ben851: Whoops, meant to add both.

PR #911, opened 2023-09-08
More quicksight datasets

sastels:
# Summary | Résumé

- add a new QuickSight user group for data set viewers (as opposed to owners)
- add more datasets!


sastels:
Note that the user group `aws_quicksight_group.dataset_full` will be deleted in order to rename the group and resource to the more accurate `aws_quicksight_group.dataset_owner`

PR #915, opened 2023-09-11
turn on quicksight again and fix some ids

sastels:
# Summary | Résumé

We removed quicksight from terraform plan and apply since the broken dataset was breaking terraform 😞 

After this dataset is deleted by AWS (support ticket) we should be able to put it back in.

This PR also fixes a couple ids that ar enot allowed to contain underscores.

NOTE: don't merge until 
- AWS deletes the dataset
- we add quicksight back in to the plan


sastels:
fixed a different way

PR #916, opened 2023-09-12
Removing secondary target groups

ben851:
# Summary | Résumé

Removing unused secondary target groups. These were used for the internal ALB migration which has been completed.

- Tested in scratch
- Smoke test in staging after apply
- Admin test in staging after apply


jimleroyer (approved):
:shipit:

PR #920, opened 2023-09-18
Creating secondary node group for prod node migration

ben851:
# Summary | Résumé

This is the first step at getting the new celery HPA tuning into prod for emails. We need to change the node type in production, which means we need to do a node migration. 

[We are following the node migration guide here](https://github.com/cds-snc/notification-terraform/blob/main/docs/nodeUpgrade.md)

- Verify the terraform plan shows the creation of the secondary node group, with no changes to primary.
- Smoke test production after apply

### DIFF
--- a/env/production/eks/terragrunt.hcl
+++ b/env/production/eks/terragrunt.hcl
@@ -34,11 +34,11 @@ include {
 }
 
 inputs = {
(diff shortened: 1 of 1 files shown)


jimleroyer (approved):
:shipit:

PR #937, opened 2023-09-22
get quicksight going again with a different template dataset

sastels:
# Summary | Résumé

Hopefully avoid the broken `templates` dataset and also fix a problem with underscores in a few `physical_table_map_id`s.

### DIFF
--- a/aws/quicksight/dataset_notification_history.tf
+++ b/aws/quicksight/dataset_notification_history.tf
@@ -6,7 +6,7 @@ resource "aws_quicksight_data_set" "notification_history" {
   import_mode = "SPICE"
 
   physical_table_map {
-    physical_table_map_id = "notification_history"
+    physical_table_map_id = "notification-history"
     relational_table {
       data_source_arn = aws_quicksight_data_source.rds.arn
       name            = "notification_history"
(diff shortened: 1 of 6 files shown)


aws/quicksight/dataset_notification_history.tf
```diff
@@ -6,7 +6,7 @@ resource "aws_quicksight_data_set" "notification_history" {
   import_mode = "SPICE"
 
   physical_table_map {
-    physical_table_map_id = "notification_history"
```
sastels, line 9: underscores not allowed here 😞

aws/quicksight/dataset_templates.tf
```diff
@@ -1,12 +1,12 @@
 # valid column types are [STRING INTEGER DECIMAL DATETIME BIT BOOLEAN JSON]
 
 resource "aws_quicksight_data_set" "templates" {
-  data_set_id = "templates"
```
sastels, line 4: the dataset with id `templates` is horribly broken in AWS and can't be deleted. AWS is looking into it, but for now let's just use a different id

ben851 (approved):
LGTM!

aws/quicksight/dataset_templates.tf
jimleroyer: That's bad, would we have to nuke it all and redo it to fix it?

PR #944, opened 2023-09-26
Parameterizing database name to account for some issues in scratch

ben851:
# Summary | Résumé

This is just a quick PR to parameterize the database name, something I require when working on scratch account. There should be no effect on staging or production.

TF Plan on staging and prod should be empty for RDS.

### DIFF
--- a/env/scratch/rds/terragrunt.hcl
+++ b/env/scratch/rds/terragrunt.hcl
@@ -41,6 +41,7 @@ inputs = {
   rds_instance_type         = "db.t3.medium"
   vpc_private_subnets       = dependency.common.outputs.vpc_private_subnets
   sns_alert_general_arn     = dependency.common.outputs.sns_alert_general_arn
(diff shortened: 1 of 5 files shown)


env/scratch/rds/terragrunt.hcl
```diff
@@ -41,6 +41,7 @@ inputs = {
   rds_instance_type         = "db.t3.medium"
   vpc_private_subnets       = dependency.common.outputs.vpc_private_subnets
   sns_alert_general_arn     = dependency.common.outputs.sns_alert_general_arn
+  rds_database_name         = "notificationcanadacascratch"
```
sastels, line 44: Did you mean to lower-case it all?

ben851: Yeah, and this is the reason I'm merging this... for some reason (and it's a bug I need to look into) this is the only way it works in scratch. At any rate, I don't see the harm in parameterizing it.

PR #947, opened 2023-09-26
join Quicksight notification_history table to services

sastels:
# Summary | Résumé

add service info to the notification_history table. If this works then I'll join in templates as well as join templates and services with the notifications table.


aws/quicksight/dataset_notifications.tf
```diff
@@ -50,6 +58,14 @@ resource "aws_quicksight_data_set" "notifications" {
...
         type = "DATETIME"
       }
+      input_columns {
+        name = "feedback_type"
+        type = "STRING"
+      }
+      input_columns {
+        name = "feedback_subtype"
+        type = "STRING"
+      }
```
sastels, line 68: Added these to notification_history so added them here too so I don't forget later

aws/quicksight/dataset_notification_history.tf
```diff
@@ -38,6 +38,14 @@ resource "aws_quicksight_data_set" "notification_history" {
         name = "notification_status"
         type = "STRING"
       }
+      input_columns {
+        name = "queue_name"
+        type = "STRING"
+      }
```
jimleroyer, line 44: Hmm the queue_name is available in the notification_history data set when I logged into Quicksights staging account. It was retrieved automatically before I assume?

jimleroyer (approved):
LGTM!

aws/quicksight/dataset_notification_history.tf
sastels: that is interesting... possibly when I was messing around with joins in the console I reimported things 🤔

PR #959, opened 2023-10-05
Setting desired node count in prod to 4

ben851:
# Summary | Résumé

Celery is now running on karpenter nodes in production. Reducing the node count for the on demand nodes to 4. This is a conservative reduction, if this behaves well, we will further reduce to 3 nodes like we have in staging.

- Smoke test prod
- Verify all pods are deployed
- Verify resource usage on nodes


sastels (approved):
📈 📉

PR #962, opened 2023-10-10
reorg alarms and add new ones for new email queues

sastels:
# Summary | Résumé

New queues for `deliver_email` are added in [this PR](https://github.com/cds-snc/notification-api/pull/1997). Here we:
- separate sms and email alarms to their own tf files.
- add alarms for new `send-email-low`, `send-email-medium`, and `send-email-high` queues.


aws/common/cloudwatch_alarms_email.tf
```diff
@@ -0,0 +1,258 @@
...
+  period                    = 60 * 60 * 12
+  statistic                 = "Average"
+  threshold                 = 0.4 / 100
+  alarm_actions             = [aws_sns_topic.notification-canada-ca-alert-critical.arn]
+  ok_actions                = [aws_sns_topic.notification-canada-ca-alert-ok.arn]
+  insufficient_data_actions = [aws_sns_topic.notification-canada-ca-alert-warning.arn]
+  treat_missing_data        = "notBreaching"
+}
+
+resource "aws_cloudwatch_metric_alarm" "sqs-email-queue-delay-warning" {
```
jimleroyer, line 74: That is wise to  keep the old queue monitoring in place for the time being. Can we add a comment please to explain it's the old one and planned to be deprecated soon enough? 🤞

sastels: good idea! And I'll add a card to check in 2 weeks if we can cleanup this code

ben851 (approved):
Eventually I'd like to explicitly create all queues in TF but this is good

PR #964, opened 2023-10-11
Adding a general system overview dashboard

ben851:
# Summary | Résumé

Adding a dashboard that provides an overall system overview with some key information that I find myself looking for often (celery errors etc)

Dashboard should deploy to staging and work correctly.

### DIFF
--- a/aws/common/dashboards.tf
+++ b/aws/common/dashboards.tf
@@ -12,7 +12,7 @@ resource "aws_cloudwatch_dashboard" "notify_system" {
             "type": "log",
             "properties": {
                 "query": "SOURCE '/aws/containerinsights/notification-canada-ca-staging-eks-cluster/application' | fields kubernetes.namespace_name as Namespace\n| stats count(*) by Namespace\n| display Namespace\n",
-                "region": "ca-central-1",
+                "region": "${var.region}",
                 "stacked": false,
                 "title": "Namespaces",
                 "view": "table"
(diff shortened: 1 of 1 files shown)


sastels:
Do you have a link to an existing dashboard?

ben851:
> Do you have a link to an existing dashboard you've been using?

Sure here it is in staging:
https://ca-central-1.console.aws.amazon.com/cloudwatch/home?region=ca-central-1#dashboards/dashboard/Ben-Notify-System

sastels:
This is nice!
A couple thoughts
- can we make the send rates per minute rather than per 5 minutes (and add text (title maybe?) to reflect that)
- this seems to be a k8s dashboard essentially. Which would be good. We have [this one](https://ca-central-1.console.aws.amazon.com/cloudwatch/home?region=ca-central-1#dashboards/dashboard/Kubernetes) already which... isn't great. Maybe we could replace that one with this?

sastels:
hrm, api lambda is on there to though, so maybe this really could be considered a system status dashboard.

ben851:
> This is nice! A couple thoughts
> 
> * can we make the send rates per minute rather than per 5 minutes (and add text (title maybe?) to reflect that)
> * this seems to be a k8s dashboard essentially. Which would be good. We have [this one](https://ca-central-1.console.aws.amazon.com/cloudwatch/home?region=ca-central-1#dashboards/dashboard/Kubernetes) already which... isn't great. Maybe we could replace that one with this?

Made the metrics per minute as requested. The other idea, you answered your own question in your other comment :)

sastels (approved):
:shipit:

aws/common/dashboards.tf
```diff
@@ -1,3 +1,338 @@
+resource "aws_cloudwatch_dashboard" "notify_system" {
+  count          = var.cloudwatch_enabled ? 1 : 0
+  dashboard_name = "Notify-System-Overview"
```
jimleroyer, line 3: Is this dashboard specific to Kubernetes?

PR #965, opened 2023-10-11
Upping vpc cni version in staging, setting prod to k8s 1.28

ben851:
# Summary | Résumé

Setting K8s version to 1.28 in prod, with the CNI version to latest (we pre-emptively upgraded the CNI in prod)

Also upping CNI version to latest in staging.

Smoke test prod
Verify you can still log into admin


jimleroyer (approved):
:shipit: 

(ideally staging and prod changes might benefit from 2 different PRs so that quick reverts work better, but I wonder how much we can really revert these once upgraded?)

PR #970, opened 2023-10-11
Dashboard fix take 2

ben851:
# Summary | Résumé

Setting what is hopefully the correct formatting of the parameterization on k8s cluster name


sastels (approved):
aren't dashboards for two accounts fun?

PR #979, opened 2023-10-19
Provider init done for prod

ben851:
# Summary | Résumé

Updating the lock files in prod so we can continue with the aws provider update.

TG Plan in the release PR should work.


jimleroyer (approved):
Ben and I talked about this change whereas he explained his reasoning, as this is a one time operation that might not be worth to include in our CI/CD automation.

PR #981, opened 2023-10-23
Reducing EKS minimum node count to 3; Setting desired count to 4

ben851:
# Summary | Résumé

Celery has been running stable in prod for a week on karpenter nodes. We can now reduce the primary nodes. This will reduce to 4, which is a conservative reduction. We will monitor the system and reduce to 3 nodes in the future.

- TF Plan in prod shows node reduction
- Smoke Test
- Verify admin works

### DIFF
--- a/env/production/eks/terragrunt.hcl
+++ b/env/production/eks/terragrunt.hcl
@@ -34,12 +34,12 @@ include {
 }
 
 inputs = {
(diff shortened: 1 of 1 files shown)


sastels (approved):
LGTM - though I think the real test will be when we allow the number of celery pods to fluctuate

PR #983, opened 2023-10-24
Revert "Try join again"

sastels:
Reverts cds-snc/notification-terraform#961

The dataset is there, but trying to import it fails. Seems to be a problem reding the status. `plan` fails too when running locally. Reverting for now...


sastels:
Removed quicksight from CI temporarily, will try to fix using local / dev

PR #957, opened 2023-10-05
Beginning Of Notify Dev

ben851:
# Summary | Résumé

Notify Dev completely deployed and up to date with main!

PR #990, opened 2023-10-26
Release 2.3.10

andrewleith:
# Summary | Résumé

- https://github.com/cds-snc/notification-terraform/pull/957
- https://github.com/cds-snc/notification-terraform/pull/962
- https://github.com/cds-snc/notification-terraform/pull/989


sastels (approved):

sastels:
hrm, why are we adding. null_resource.aws_shield_subscription?

ben851:
> hrm, why are we adding. null_resource.aws_shield_subscription?

This is for the new environment creation. It will run once here, fail successfully (since the shield is already enabled here),  and shouldn't come back.

PR #995, opened 2023-10-31
Removed broken quicksight dataset in aws and tfstate. Plan works locally

ben851:
# Summary | Résumé

The quicksight plan and deploy are re-enabled and should work in staging.

# Test instructions | Instructions pour tester la modification

Merge to main - deploy to staging works.

PR #996, opened 2023-10-31
Revert "Removed broken quicksight dataset in aws and tfstate. Plan works locally"

ben851:
Reverts cds-snc/notification-terraform#995

The plan works but apply still fails


sastels (approved):
😢

PR #997, opened 2023-10-31
Cloudfront s3

ben851:
# Summary | Résumé

Adding a hack to get the cloudfront assets for admin into the S3 bucket programmatically.

I tested in dev that it will overwrite existing files automatically and thus we shouldn't have any conflicts with existing resources in staging and prod.

Terragrunt deploy
Ensure that admin is still working as expected

### DIFF
--- a/aws/common/uploadS3Assets.tf
+++ b/aws/common/uploadS3Assets.tf
@@ -1,0 +1,7 @@
+resource "aws_s3_object" "assets" {
+
+    bucket = aws_s3_bucket.asset_bucket.id
+    for_each = fileset("/var/tmp/notification-admin/app/assets/cloudfront","**")
+    key = each.value
+    source = "/var/tmp/notification-admin/app/assets/cloudfront/${each.value}"
+}
--- a/env/production/common/terragrunt.hcl
+++ b/env/production/common/terragrunt.hcl
@@ -2,6 +2,19 @@
 #
 terraform {
   source = "git::https://github.com/cds-snc/notification-terraform//aws/common?ref=v${get_env("INFRASTRUCTURE_VERSION")}"
+  
+  before_hook "get-admin" {
+    commands     = ["apply", "plan"]
+    execute      = ["git", "clone","-b", "asset-update", "https://github.com/cds-snc/notification-admin.git", "/var/tmp/notification-admin"]
+    run_on_error = true
+
+  }
+
+  after_hook "cleanup-admin" {
+    commands     = ["apply", "plan"]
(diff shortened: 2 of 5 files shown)


env/production/common/terragrunt.hcl
```diff
@@ -2,6 +2,19 @@
 #
 terraform {
   source = "git::https://github.com/cds-snc/notification-terraform//aws/common?ref=v${get_env("INFRASTRUCTURE_VERSION")}"
+  
+  before_hook "get-admin" {
+    commands     = ["apply", "plan"]
+    execute      = ["git", "clone","-b", "asset-update", "https://github.com/cds-snc/notification-admin.git", "/var/tmp/notification-admin"]
```
CalvinRodo, line 8: Should this be pulling down the same version of prod assets as the "INFRASTRUCTURE_VERSION" or should it always get the HEAD of the main branch in the notification-admin repo.

CalvinRodo (commented):
see my comment about always getting the latest version of the main branch for notification-admin

env/production/common/terragrunt.hcl
ben851: Whoops I clearly missed updating this from testing. 

I think it should be getting latest from main if for no other reason than there being no tagging done on the admin branch, making it hard to do anything else.

aws/common/uploadS3Assets.tf
```diff
@@ -0,0 +1,7 @@
+resource "aws_s3_object" "assets" {
+
```
jimleroyer, line 2: Can we add comments all around to tell of this hack? It's really the most important aspect of comments, i.e. explain hacks. 😉

env/production/common/terragrunt.hcl
jimleroyer: It's an interesting issue that could lead to some edge cases.

aws/common/uploadS3Assets.tf
```diff
@@ -0,0 +1,10 @@
+# This is a bit of a hack in that we're cloning an external repo into a temp directory so that we can populate cloudfront.
+# We will have to review a better way of doing this in the future, but this will get us going in the right direction
```
jimleroyer, line 2:
```suggestion
# This is hack'ish as we're cloning an external repo into a temp directory so that we can populate CloudFront.
# We will have to review a better way of doing this in the future, but our long term goal of having one source of
# truth for the assets is accomplished.
```

jimleroyer (approved):
This is fantastic, life is plastic. :shipit:

PR #1002, opened 2023-11-02
chore: remove legacy TLS ALB listener

patheard:
# Summary
Remove the port 4444 legacy TLS load balancer listener and its associated security group ingress rule.

This listener was set to return a `403 Forbidden` to all requests back in November 2021, so the risk of removing it is low.

The reason for the removal is that it is being flagged as a Security Hub finding for `0.0.0.0/0` ingress on an unexpected port.

# Related
- #262 
- #321
- https://github.com/cds-snc/platform-core-services/issues/471

### DIFF
--- a/aws/eks/alb.tf
+++ b/aws/eks/alb.tf
@@ -66,27 +66,6 @@ resource "aws_lb_listener" "notification-canada-ca-80" {
   }
 }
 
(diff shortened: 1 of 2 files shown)


ben851 (approved):
LGTM!

PR #1003, opened 2023-11-02
Rudementary karpenter error check

ben851:
# Summary | Résumé

This will alert us if we see that a pod was unable to be scheduled.

We can break karpenter in staging by setting an invalid node selector.


aws/eks/cloudwatch_alarms.tf
```diff
@@ -729,4 +729,20 @@ resource "aws_cloudwatch_metric_alarm" "documentation-evicted-pods" {
...
+  comparison_operator = "GreaterThanOrEqualToThreshold"
+  evaluation_periods  = "1"
+  metric_name         = aws_cloudwatch_log_metric_filter.celery-error[0].metric_transformation[0].name
+  namespace           = aws_cloudwatch_log_metric_filter.celery-error[0].metric_transformation[0].namespace
+  period              = 120
+  statistic           = "Sum"
+  threshold           = 1
+  treat_missing_data  = "notBreaching"
+  #Warning for now
+  alarm_actions = [var.sns_alert_warning_arn]
```
jimleroyer, line 747: The alarm name is set top critical but we sent to the warning SNS: is that intended?

ben851: Yes, just for testing... I don't want to spam us. If after QA it's good we'll up it to critical

jimleroyer (approved):

PR #1005, opened 2023-11-03
excplicitly setting image mime type

ben851:
# Summary | Résumé

We're running into image mime type issues, this should explicitly set mime type to images.

We will need to invalidate the cloudfront cache afterwards.

Apply staging
Invalidate cloudfront cache
Navigate admin and verify images work in staging


jzbahrai (approved):

sastels (approved):
Went through the plan and verified that all the files that we're setting the content type for are svgs

PR #1011, opened 2023-11-07
upping the evaluation period threshold on evicted pod alarms

ben851:
# Summary | Résumé

Upping the alarm evaluation period threshold on evicted pod alarms to 3 consecutive periods. Evicted pods are now a possible valid state since we are using karpenter to reconcile nodes. 3 periods/3 minutes should be enough time for the system to do a full recycle without triggering alarms.

Verify that we do not receive alarms when karpenter reconciles nodes.


jimleroyer (approved):
LGTM

PR #1015, opened 2023-11-09
Add the `Cross-Origin-Resource-Policy' HTTP header

andrewleith:
# Summary | Résumé

This PR adds the `Cross-Origin-Resource-Policy` HTTP header to the `204` response of requests denied by the `valid_paths` rule in our WAF.  

## Context

- This issue (Insufficient Site Isolation Against Spectre Vulnerability) was identified by an OWASP scan
- While there likely isn't an issue with the way the WAF is returning the request (especially considering it isn't our app doing it) it can't hurt to add this header and get rid of the OWASP issue.

## Other reading
- [Details on the CORP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cross-Origin_Resource_Policy)


aws/eks/waf.tf
```diff
@@ -283,6 +283,10 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
             name  = "Strict-Transport-Security"
             value = "max-age=63072000; includeSubDomains; preload"
           }
+          response_header {
+            name  = "Cross-Origin-Resource-Policy"
+            value = "same-origin"
```
jimleroyer, line 288: Very good! thanks!

PR #1016, opened 2023-11-09
Alarm to notify team if there was a failed login of more than 10 times

jzbahrai:
# Summary | Résumé

Add an alarm to check if someone has logged in more than 10 times

### DIFF
--- a/aws/eks/cloudwatch_alarms.tf
+++ b/aws/eks/cloudwatch_alarms.tf
@@ -715,3 +715,17 @@ resource "aws_cloudwatch_metric_alarm" "documentation-evicted-pods" {
   ok_actions                = [var.sns_alert_warning_arn]
   insufficient_data_actions = [var.sns_alert_warning_arn]
 }
+
+resource "aws_cloudwatch_metric_alarm" "failed-login-count-5-minute-warning" {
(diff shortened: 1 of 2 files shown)


aws/eks/cloudwatch_alarms.tf
```diff
@@ -715,3 +715,17 @@ resource "aws_cloudwatch_metric_alarm" "documentation-evicted-pods" {
   ok_actions                = [var.sns_alert_warning_arn]
   insufficient_data_actions = [var.sns_alert_warning_arn]
 }
+
+resource "aws_cloudwatch_metric_alarm" "failed-login-count-5-minute-warning" {
+  alarm_name          = "failed-login-count-5-minute-warning"
+  alarm_description   = "One user had a failed login count of more than 10 times in 5 minutes"
```
whabanks, line 721: Are we counting per user or just 10 failed logins over 5 minutes in general?

whabanks (approved):
LGTM!

aws/eks/cloudwatch_alarms.tf
jzbahrai: its per user. this will only show up once as in the same user id will not be allowed to even try loggin in past 10 attempts without contacting us

PR #1017, opened 2023-11-09
fix: centralize workflow failure notifications

patheard:
# Summary
Add a workflow that is triggered when named workflow runs are completed.  If these workflows have a status of `failure` a message will be posted to Slack with a link to the failed run.


ben851 (approved):
To the moon!

PR #1023, opened 2023-11-14
fix: remove RDS version constraint on update check

patheard:
# Summary
Update the RDS update check script to remove the `engine-version` parameter as this was causing an empty array of versions to be returned.

Output of the script will be the same with the highest available Postgres version available:
```
11.21, 12.9, 13.9, 14.9, 15.4
```


jimleroyer (approved):
:shipit:

PR #1026, opened 2023-11-14
stop qs plan and apply again

sastels:
# Summary | Résumé

😭


sastels:
reverted offending PRs instead

PR #1033, opened 2023-11-16
fix qs on dev

sastels:
# Summary | Résumé

changes we just made to get QuickSight working in the dev environment.


aws/quicksight/subscription.tf
```diff
@@ -1,5 +1,5 @@
 resource "aws_quicksight_account_subscription" "subscription" {
-  account_name          = "quicksight-test-${var.env}"
+  account_name          = "quicksight-test-${var.env}-${var.account_id}"
```
sastels, line 2: This change will cause our current staging quicksight to be deleted and recreated 😞 - you can't change the name of the subscription.

That's ok, but maybe we should decide for good what the name should be. probably should get rid of the "test" in there 😅

sastels:
Note that to allow a QuickSight subscription to be deleted (ie in this case to change the name) we must run
```
aws quicksight update-account-settings --no-termination-protection-enabled --aws-account-id << ID >> --default-namespace default --region=ca-central-1
```

jimleroyer (approved):

aws/quicksight/subscription.tf
sastels: got rid of "test" in the PR, and tested on dev that we can
- do the apply, which recreates an (empty) subscription
- do the apply again, which puts all the stuff in it (user groups, datasets, etc)

PR #1039, opened 2023-11-20
Bring throttled SMS delay alarms in line with current SLOs

sastels:
# Summary | Résumé

throttled SMS are slow, alarms go off whenever a service tries to send a lot. We adjusted our SLOs for these, and this PR adjusts the alarms accordingly.


jimleroyer (approved):
LGTM! thanks!

PR #1045, opened 2023-11-22
Removing the limit on celery errors for notify system overview

ben851:
# Summary | Résumé

Removing the limit 20 on the celery error query so that we can see all of the error messages for a given time period.

Tested deployment works in dev

- Staging apply
- Pick a long time window in the notify-system-overview dashboard and make sure you see more than 20 errors.


sastels (approved):
👍 Note that the default limit for cloudwatch queries is 1000, which is (hopefully!) fine

jimleroyer (approved):

PR #1053, opened 2023-11-23
fix: file scanning target account

patheard:
# Summary
Update the `s3_scan_object` module configuration so that all non-production Notify accounts use the Scan Files Staging service.

This will stop pentesting and load testing in the Notify Staging environment from impacting the Scan Files Production service.

# ⚠️  Note
Once this merges in Staging, the [`sqs_event_accounts` input will need to be updated in Scan Files](https://github.com/cds-snc/scan-files/blob/dfc80acd4a5a85cac7b79505be091618aa0370a9/terragrunt/env/staging/s3_scan_object/terragrunt.hcl#L27).

### DIFF
--- a/aws/common/file_scanning.tf
+++ b/aws/common/file_scanning.tf
@@ -1,6 +1,13 @@
+locals {
+  scan_files_account = var.env == "production" ? "806545929748" : "127893201980"
+}
+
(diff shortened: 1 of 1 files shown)


ben851 (approved):
LGTM

PR #1061, opened 2023-11-30
Reducing prod to 4 nodes to allow scaling to do its thing again

ben851:
# Summary | Résumé

Reduce prod to 4 primary nodes so that we delegate most celery instances to spot nodes.

Smoke test in production
Verify logging
Verify no errors in notification-ops
Verify 4 on demand nodes 
```
kubectl describe nodes | grep ON_DEMAND
```
Verify there are spot nodes
```
kubectl describe nodes | grep spot
```
Use K9s to validate that all deployments are able to spin up to their desired count

### DIFF
--- a/env/production/eks/terragrunt.hcl
+++ b/env/production/eks/terragrunt.hcl
@@ -40,12 +40,12 @@ include {
 }
 
 inputs = {
-  primary_worker_desired_size               = 8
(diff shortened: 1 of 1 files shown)


jimleroyer (approved):
LGTM

PR #1062, opened 2023-12-01
Join notifications with cloud formation

sastels:
# Summary | Résumé

Join notifications-services-orgs-templates in cloudformation. Also pulls the notifications table from RDS rather than using an already-pulled notifications data set.

This will allow us to remove the click-ops joined version.

Play with table in QuickSight!


ben851 (approved):
LGTM!

PR #1066, opened 2023-12-04
add joined notifications_history table

sastels:
# Summary | Résumé

Similar to #1062 join the notification_history table in cloud formation.

tested on dev

Note: before releasing to prod we should delete the existing notification_history_clickops table to free SPICE capacity.


aws/quicksight/dataset_notifications.tf
```diff
@@ -54,10 +54,6 @@ resource "aws_cloudformation_stack" "notifications" {
                     Name = "id",
                     Type = "STRING"
                   },
-                  {
-                    Name = "to",
-                    Type = "STRING"
-                  },
```
sastels, line 60: This wasn't in the dataset output, but there's no reason to even read it in in the first place.

ben851 (approved):

PR #1069, opened 2023-12-07
feat: add client VPN for private subnet access

patheard:
# Summary
Add a client VPN that allows access to the private subnets of the VPC.  Only authorized users will be allowed to connect, which is controlled through IAM Identity Center group membership.

Authentication is performed using the existing AWS SSO identity provider.

# Related
- https://github.com/cds-snc/platform-core-services/issues/508


env/production/eks/terragrunt.hcl
```diff
@@ -28,6 +28,7 @@ dependency "common" {
     sqs_send_sms_low_queue_name               = ""
     sqs_send_sms_medium_queue_name            = ""
     sqs_send_sms_high_queue_name              = ""
+    client_vpn_security_group_id              = "sg-1234"
```
ben851, line 31: Is this a common CDS security group?

aws/common/client_vpn.tf
```diff
@@ -0,0 +1,40 @@
...
+  client_vpn_self_service_saml_metadata_document = var.env == "production" ? var.client_vpn_self_service_saml_metadata : null
+  client_vpn_saml_metadata_document              = var.client_vpn_saml_metadata
+
+  billing_tag_value = "notification-canada-ca-${var.env}"
+}
+
+#
+# Certificate used for VPN communication
+#
+resource "aws_acm_certificate" "client_vpn" {
```
ben851, line 28: How is this cert generated? Do we have docs on this?  Can we get TF to generate the cert itself?

env/production/eks/terragrunt.hcl
patheard: No, it's just a dummy value so that I would see in the TF plan the mock value was being used.  Happy to change this to another value.

aws/common/client_vpn.tf
patheard: I generated it using the [`hashicorp/tls` provider](https://registry.terraform.io/providers/hashicorp/tls/latest/docs) locally and then moved the private key and cert into a secret.  We could absolutely use this provider directly with the only catch being that the [private key would be stored plain text in the TF state](https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/private_key).  

Probably not a big deal in our case because of how our state access is restricted and what we're using this cert for.  What do you think?

ben851: Yeah, I get that it sucks that it's stored in plain text... but the s3 is at least encrypted at rest and it's private... for me this is enough, and ensures that we can continue to deploy with few out of band steps

sastels:
Not sure exactly why / how that change to the QuickSight dataset got there but don't need (not use) that updated_at field 🤷

ben851:
I screwed up this branch. Closing in favour of #1098

PR #1070, opened 2023-12-07
add primary and scalable pods to system overview dashboard

sastels:
# Summary | Résumé

add new primary / scalable celery deployments to dashboard. Should look like [this](https://ca-central-1.console.aws.amazon.com/cloudwatch/home?region=ca-central-1#dashboards/dashboard/Notify-System-Overview-SJA) (in prod):
image

### DIFF
--- a/aws/eks/dashboards.tf
+++ b/aws/eks/dashboards.tf
@@ -7,19 +7,18 @@ resource "aws_cloudwatch_dashboard" "notify_system" {
         {
             "height": 6,
             "width": 6,
-            "y": 17,
-            "x": 11,
+            "y": 3,
+            "x": 18,
             "type": "metric",
             "properties": {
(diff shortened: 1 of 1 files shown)


aws/eks/dashboards.tf
```diff
@@ -380,20 +380,6 @@ resource "aws_cloudwatch_dashboard" "notify_system" {
                 "sparkline": true
             }
         },
-        {
```
ben851, line 383: We're getting rid of errors?

ben851 (approved):

aws/eks/dashboards.tf
sastels: We still have the list of all celery errors at the bottom, I figured that was fine. Thoughts?

PR #1071, opened 2023-12-07
Add login_events QuickSight dataset

sastels:
# Summary | Résumé

Add login_events dataset to QuickSight along with a refresh schedule.

tested on dev


jzbahrai (approved):

jimleroyer (changes requested):
Did we ask to policies if location of users is too sensitive or if it would be ok within QuickSight?

jimleroyer (approved):
The `data`  column that might contain sensitive user information is not explicitly imported so we're good to go. 🚀

PR #1072, opened 2023-12-07
set prod to 5 on demand nodes

ben851:
# Summary | Résumé

Upping prod to 5 nodes to accomodate all the primary celeries!

Prod should show 5 on demand nodes when complete.


sastels (approved):
💸

PR #1076, opened 2023-12-12
Added rds to system status lambda

jzbahrai:
# Summary | Résumé

Add rds to system status lambda


jzbahrai:
If you look at the plan, it indicated that we are deleting the non-vpc policies :
I double checked with pat and that is okay. We are adding an ecr security group and that will cover all the cloudwatch policies that we are deleting.

sastels (approved):
LGTM! Let's see what happens!

PR #1084, opened 2023-12-28
Setting service bounce rate alarm to 12 hour period

ben851:
# Summary | Résumé

Reducing the frequency of the bounce rate alert from one minute to 12 hours.

Terraform apply in staging
Run a test with lots of bounces
See if we get alerted after 12 hours


whabanks (approved):
LGTM!

PR #1114, opened 2024-01-11
Adding in the logical replication parameter to rds

ben851:
# Summary | Résumé

This will add the logical replication parameter to the RDS cluster in preparation for the DB major upgrade.

I've also updated the dev rds terragrunt.hcl 

Apply to staging, proceed with database migration


aws/rds/rds.tf
```diff
@@ -58,6 +58,12 @@ resource "aws_rds_cluster_parameter_group" "default" {
     value = "ddl"
   }
 
+  parameter {
+    name         = "rds.logical_replication"
+    value        = "1"
+    apply_method = "pending-reboot"
+  }
+
```
jimleroyer, line 66: What would be the database that would subscribe to the replication? Upcoming in future PRs?

jimleroyer (approved):

aws/rds/rds.tf
ben851: Yep exactly, this is needed to prep for the blue/green deployment, but as you can see, this parameter requires a restart (aka downtime) to enable. If we get this in now, we can kill two birds with one stone when we do the minor version upgrade as well

PR #1115, opened 2024-01-11
Add alias_name for lambda functions

jzbahrai:
# Summary | Résumé

Add alias_name for lambda functions


sastels (approved):
Nice!

As an aside, noticed that we're doing the api lambda manually (ie not using a module like this!). Might want to think about switching to this module at some point!

PR #1119, opened 2024-01-15
Adding force upgrade option for k8s node ugrades

ben851:
# Summary | Résumé

This will provide us the option to force upgrades in the event that pods fail to evict. This option will be disabled by default and will have to be manually enabled should we ever need to use it. 

I also changed the nodeUpgrade variable to node_upgrade to keep with the variable formatting.

Terraform apply works, set force_upgrade to true, apply again, proceed with node upgrade.


sastels (approved):

jimleroyer (approved):
That's great.

PR #1120, opened 2024-01-16
fix: update RDS version script with engine-version

patheard:
# Summary
Update the script that checks for available RDS cluster version updates to include the cluster's current engine version.

This will prevent versions that are lower than the clusters current engine version from being displayed.


ben851 (approved):
LGTM

PR #1122, opened 2024-01-16
Upgrading lower states to latest

ben851:
# Summary | Résumé

This will upgrade non-prod k8 worker nodes to 1.28.5-20240110

TF Apply in staging - verify the force evict pods is not required.

Smoke test staging


jimleroyer (approved):
LGTM

PR #1124, opened 2024-01-17
Adding logic and readme to dev environment to help keep things in sync

ben851:
# Summary | Résumé

This will let you sync dev.tfvars from one password, and any changes you don't have.

Install the 1Password CLI tool as per the readme, do a TF apply/plan.


env/terragrunt.hcl
```diff
@@ -25,11 +26,21 @@ inputs = {
   cbs_satellite_bucket_name = "cbs-satellite-${local.vars.inputs.account_id}"
 }
 
+terraform {
+
+  before_hook "before_hook" {
+    commands     = local.vars.inputs.env == "dev" ? ["apply", "plan"] : []
```
sastels, line 32: what about init?

scripts/checkEnvFile.sh
```diff
@@ -0,0 +1,49 @@
...
+  echo "TFVars file doesn't exist... fetching."
+  op read op://4eyyuwddp6w4vxlabrr2i2duxm/"TFVars - Dev"/notesPlain > $VARFILE
+  echo "Done."
+  echo "STOPPING THIS RUN. TERRAFORM NEEDS TO BE RE-RUN TO PICK UP THE DEV FILE."
+  echo "PLEASE RE-RUN YOUR COMMAND."
+  exit 1
+else
+  echo "TFVars file exists."
+  CHECK=true
+fi
```
sastels, line 15: something went wrong the first time with my auth and I ended up with an empty dev.tfvars file - and then the second step asked me if I wanted to put that into 1Password 😱  - what about also fetching if the file is empty?

sastels (changes requested):
just a couple questions / suggestions. Other than that, worked as advertised 🎉

sastels (approved):
So very cool. Tested locally (I mean, where else would you test this?)

PR #1125, opened 2024-01-17
Blazer db instance

ben851:
# Summary | Résumé

This creates a new DB instance outside of the main instance group but within the same database cluster. 

This DB instance has its own endpoint, which is then fed to blazer in database-tools for consumption.

Notes:
- Originally I wanted to create a complete separate cluster with replication enabled, but this doesn't seem to be supported for aurora databases
- I also wanted to put in a secondary read only proxy that went to this instance specifically, but that also is not supported
- When we upgrade to Postgres 15 we can convert this new instance into a serverless instance to save money. I have left the config for this in this PR but commented out.

TF Apply in staging, connect to blazer, verify the endpoint is NOT the RDS proxy, run queries.


jzbahrai (approved):
lets give it a try!

PR #1123, opened 2024-01-17
feat: add alarm and log metric for queues not being active

andrewleith:
# Summary | Résumé

This PR adds a metric filter to check the queues are being emptied, i.e. the `beat_inbox_*` methods in [scheduled_tasks.py](https://github.com/cds-snc/notification-api/blob/main/app/celery/scheduled_tasks.py#L284) are firing.

It also adds 2 cloud watch alarms:
- warning if no queue activity in 1 minute
- critical if no queue activity in 5 minutes

PR #1131, opened 2024-01-23
chore: update alarm names/descriptions/metric filters to be more accu…

andrewleith:
# Summary | Résumé

This pr amends #1123 to make the alarm names, alarm description, and metric filter names more accurate, i.e. this is monitoring that the beat inbox celery tasks are constantly firing and not so much watching the queues themselves.

### DIFF
--- a/aws/eks/cloudwatch_alarms.tf
+++ b/aws/eks/cloudwatch_alarms.tf
@@ -908,14 +908,14 @@ resource "aws_cloudwatch_metric_alarm" "karpenter-replicas-unavailable" {
   }
 }
 
-resource "aws_cloudwatch_metric_alarm" "queues-not-active-1-minute-warning" {
+resource "aws_cloudwatch_metric_alarm" "sqs-beat-inbox-tasks-not-active-1-minute-warning" {
   count               = var.cloudwatch_enabled ? 1 : 0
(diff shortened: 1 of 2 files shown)


aws/eks/cloudwatch_alarms.tf
```diff
@@ -908,29 +908,29 @@ resource "aws_cloudwatch_metric_alarm" "karpenter-replicas-unavailable" {
...
+  namespace           = aws_cloudwatch_log_metric_filter.beat-inbox-tasks-are-active[0].metric_transformation[0].namespace
   period              = "60"
   statistic           = "Sum"
   threshold           = 1
   treat_missing_data  = "breaching"
   alarm_actions       = [var.sns_alert_warning_arn]
 }
 
-resource "aws_cloudwatch_metric_alarm" "queues-not-active-5-minutes-critical" {
+resource "aws_cloudwatch_metric_alarm" "beat-inbox-tasks-not-active-5-minutes-critical" {
```
jimleroyer, line 926: Can we keep the `queue` part in the name? beat and tasks are heave on Celery and could be true of other components. A name such as `aggregating-queues` would carry the meaning. We can talk offline when we have 5 minutes?

jimleroyer (approved):

PR #1132, opened 2024-01-23
Refresh all QuickSight datasets

sastels:
# Summary | Résumé

Add nightly refreshes to remaining datasets


aws/quicksight/dataset_login_events.tf
```diff
@@ -52,7 +52,7 @@ resource "aws_quicksight_refresh_schedule" "login_events" {
   depends_on  = [aws_quicksight_data_set.login_events]
 
   schedule {
-    refresh_type = "INCREMENTAL_REFRESH"
+    refresh_type = "FULL_REFRESH"
```
jimleroyer, line 55: Incaremental refresh does not work?

sastels: I think it does but I wanted to spend a little bit more time making sure parameters were chosen correctly (ie that **everything** since the last refresh would get brought in). I think that means an overlap period of some length. So for now I figured I'd just make full so that we could have confidence we're not missing anything.

aws/quicksight/dataset_notifications.tf
```diff
@@ -384,3 +384,19 @@ resource "aws_cloudformation_stack" "notifications" {
...
+  data_set_id = "notifications"
+  schedule_id = "schedule-notifications"
+  depends_on  = [aws_cloudformation_stack.notifications]
+
+  schedule {
+    refresh_type = "FULL_REFRESH"
+
+    schedule_frequency {
+      interval        = "DAILY"
+      time_of_the_day = "07:20"
```
jimleroyer, line 399: I wonder if we should refresh with a dependency order, i.e. same as of now except for notifications being to refresh.

jimleroyer (approved):
Left a few questions but Steve is the master of the universe.

aws/quicksight/dataset_notifications.tf
sastels: That's not a bad idea! my order was essentially random 😅

PR #1133, opened 2024-01-25
VPN Internal DNS

ben851:
# Summary | Résumé

This will create the internal DNS and the internal load balancer for the dev VPN. This is the pre-requisite for the manifests PR that will create a route for hasura to be reachable by our internal DNS.

Plan/Apply work in staging.


aws/eks/dns.tf
```diff
@@ -86,3 +86,26 @@ resource "aws_route53_record" "api-weighted-0-scratch-notification-A" {
...
+  }
+}
+
+resource "aws_route53_record" "wildcard_CNAME" {
+  zone_id = var.internal_dns_zone_id
+  name    = "*.${var.env}.notification.internal"
+  type    = "CNAME"
+  ttl     = "60"
+  records = ["${var.env}.notification.internal"]
+}
```
jimleroyer, line 111: Random question that is not directly related to this PR but do you know of a tool that can ingest terraform files and make a nice diagram of components along with how these are related (with connections and all)?

jimleroyer (approved):
LGTM

aws/eks/dns.tf
ben851: There's [terarform graph](https://developer.hashicorp.com/terraform/cli/commands/graph) but I've found it's usefulness limited. Definitely something to look into some more though

PR #1146, opened 2024-02-08
Adding internal HTTP ingress/egress to eks SG

ben851:
# Summary | Résumé

This allows ingress/egress for HTTP between the EKS worker nodes and the internal load balancer. External traffic is still blocked.

ALSO: Added the lock.hcl's after init --upgrade so that we don't have to do it again

Apply in staging, connect to:
https://hasura.staging.notification.internal


sastels:
Were the lock changes needed or is that just updating?

ben851:
> Were the lock changes needed or is that just extra updating?
Oop yeah they're needed since we merged that renovate PR

sastels (approved):

PR #1161, opened 2024-02-13
Setting coreDNS to run on ON_DEMAND only

ben851:
# Summary | Résumé

This will enforce coreDNS running on ON_DEMAND nodes only.

TF Apply in staging, verify that coredns is not running on spot instances

### DIFF
--- a/aws/eks/eks.tf
+++ b/aws/eks/eks.tf
@@ -161,6 +161,9 @@ resource "aws_eks_addon" "coredns" {
           loadbalance
       }
       EOF
+    "nodeSelector" : {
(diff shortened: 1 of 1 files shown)


aws/eks/eks.tf
```diff
@@ -161,6 +161,9 @@ resource "aws_eks_addon" "coredns" {
           loadbalance
       }
       EOF
+    "nodeSelector" : {
```
andrewleith, line 164: So if you don't specify, it ends up as being autoscaled?

andrewleith (approved):
LGTM

PR #1165, opened 2024-02-15
removing logging from coredns

ben851:
# Summary | Résumé

This will disable logging from coredns. To be released when we've confirmed that the coredns issue has been resolved.

```shell
kubectl rollout restart deployment/coredns -n kube-system
```
kubectl logs on the new pods, verify only the version logs are sent to stdout.


jimleroyer (approved):
LGTM

PR #1167, opened 2024-02-19
fix: remove trailing slash from FR route

andrewleith:
# Summary | Résumé

Make the route work in FR by removing trailing slash


sastels (approved):
Third time's the charm 😂

PR #1169, opened 2024-02-19
remove anything with blazer_slack in it

sastels:
# Summary | Résumé

Delete code around the blazer Slack webhook since we are not using it.


jimleroyer (approved):
LGTM 🚢

PR #1173, opened 2024-02-20
Logging SMS usage costs

jimleroyer:
# Summary | Résumé

This adds S3 buckets into which the SMS usage report, including cost, will be generated and dropped into.

Instructions on how to register to these are located here:
https://docs.aws.amazon.com/sns/latest/dg/sms_stats_usage.html

We'll have to merge to see if the permissions and config works.

### DIFF
--- a/aws/common/s3.tf
+++ b/aws/common/s3.tf
@@ -384,7 +384,7 @@ module "sns_sms_usage_report_bucket" {
   force_destroy     = var.force_destroy_s3
   billing_tag_value = "notification-canada-ca-${var.env}"
 
-  lifecycle_rule = { "lifecycle_rule" : { "enabled" : "true", "expiration" : { "days" : "90" } } }
+  lifecycle_rule = { "lifecycle_rule" : { "enabled" : "true", "expiration" : { "days" : "7" } } }
 
   tags = {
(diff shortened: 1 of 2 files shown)


aws/common/s3.tf
```diff
@@ -376,3 +376,155 @@ module "cbs_logs_bucket" {
...
 }
+
+module "sns_sms_usage_report_bucket" {
+  source = "github.com/cds-snc/terraform-modules//S3_log_bucket?ref=v6.0.3"
+
+  bucket_name       = "notification-canada-ca-${var.env}-sms-usage-logs"
+  force_destroy     = var.force_destroy_s3
+  billing_tag_value = "notification-canada-ca-${var.env}"
+
+  lifecycle_rule = { "lifecycle_rule" : { "enabled" : "true", "expiration" : { "days" : "90" } } }
```
sastels, line 387: We can merge and check this out, but I think this is going to log the destination phone # too, so we might need a shorter retention.

jimleroyer: True! I will lower it down to 7 days for now. cc @yaelberger-commits

sastels (approved):
LGTM! :shipit:

PR #1175, opened 2024-02-21
new sql based notifications dataset

sastels:
# Summary | Résumé

Use SQL to build a dataset that is the union of the notifications and notification_history tables.


sastels:
Done in #1179

PR #1178, opened 2024-02-22
Logging SMS usage costs (second attempt)

jimleroyer:
# Summary | Résumé

This adds S3 buckets into which the SMS usage report, including cost, will be generated and dropped into.

Instructions on how to register to these are located here:
https://docs.aws.amazon.com/sns/latest/dg/sms_stats_usage.html

First attempt was #1173 . This one hopefully fixes the previous attempt that did not work:

```
 "errorCode": "AuthorizationErrorException",
    "errorMessage": "You are not authorized to perform actions for the provided bucket",
```

The cause is likely that the us-west-2 provider was not defined on the new s3 bucket that was meant for that region. This new PR fixes that.

We'll have to merge to see if the permissions and config works.

### DIFF
--- a/aws/common/s3.tf
+++ b/aws/common/s3.tf
@@ -454,7 +454,9 @@ POLICY
 }
 
 module "sns_sms_usage_report_bucket_us_west_2" {
-  provider = aws.us-west-2
+  providers = {
+    aws = aws.us-west-2
+  }
 
   source = "github.com/cds-snc/terraform-modules//S3_log_bucket?ref=v6.0.3"
 
--- a/aws/common/s3.tf
+++ b/aws/common/s3.tf
@@ -509,7 +509,7 @@ resource "aws_s3_bucket_policy" "sns_sms_usage_report_bucket_us_west_2_policy" {
 					"aws:SourceAccount": "${var.account_id}"
 				},
 				"ArnLike": {
-					"aws:SourceArn": "arn:aws:sns:${var.region}:${var.account_id}:*"
+					"aws:SourceArn": "arn:aws:sns:us-west-2:${var.account_id}:*"
 				}
 			}
 		},
--- a/aws/common/s3.tf
+++ b/aws/common/s3.tf
@@ -526,7 +526,7 @@ resource "aws_s3_bucket_policy" "sns_sms_usage_report_bucket_us_west_2_policy" {
 					"aws:SourceAccount": "${var.account_id}"
 				},
 				"ArnLike": {
-					"aws:SourceArn": "arn:aws:sns:${var.region}:${var.account_id}:*"
+					"aws:SourceArn": "arn:aws:sns:us-west-2:${var.account_id}:*"
 				}
 			}
 		}
--- a/aws/common/s3.tf
+++ b/aws/common/s3.tf
@@ -376,3 +376,155 @@ module "cbs_logs_bucket" {
     CostCenter = "notification-canada-ca-${var.env}"
   }
 }
+
+module "sns_sms_usage_report_bucket" {
+  source = "github.com/cds-snc/terraform-modules//S3_log_bucket?ref=v6.0.3"
+
+  bucket_name       = "notification-canada-ca-${var.env}-sms-usage-logs"
+  force_destroy     = var.force_destroy_s3
+  billing_tag_value = "notification-canada-ca-${var.env}"
+
+  lifecycle_rule = { "lifecycle_rule" : { "enabled" : "true", "expiration" : { "days" : "7" } } }
+
+  tags = {
+    CostCenter = "notification-canada-ca-${var.env}"
+  }
+}
+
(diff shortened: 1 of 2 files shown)


aws/common/s3.tf
```diff
@@ -376,3 +376,159 @@ module "cbs_logs_bucket" {
...
+				}
+			}
+		}
+	]
+}
+POLICY
+}
+
+module "sns_sms_usage_report_bucket_us_west_2" {
+  provider = aws.us-west-2
```
jimleroyer, line 457: This was missing on the first attempt The bucket needs to be created in `us-west-2` for the `update-sms-prefs-us-west-2` for the resource to see it (also located in `us-west-2`).

aws/common/s3.tf
```diff
@@ -376,3 +376,159 @@ module "cbs_logs_bucket" {
...
+
+  lifecycle_rule = { "lifecycle_rule" : { "enabled" : "true", "expiration" : { "days" : "7" } } }
+
+  tags = {
+    CostCenter = "notification-canada-ca-${var.env}"
+  }
+}
+
+resource "aws_s3_bucket_policy" "sns_sms_usage_report_bucket_us_west_2_policy" {
+  provider = aws.us-west-2
```
jimleroyer, line 473: This was missing on the first attempt The bucket needs to be created in `us-west-2` for the `update-sms-prefs-us-west-2` for the resource to see it (also located in `us-west-2`).

patheard: Same comment about this provider change likely not being needed.

aws/common/s3.tf
```diff
@@ -376,3 +376,159 @@ module "cbs_logs_bucket" {
...
+				"Service": "sns.amazonaws.com"
+			},
+			"Action": "s3:PutObject",
+			"Resource": "arn:aws:s3:::${module.sns_sms_usage_report_bucket_us_west_2.s3_bucket_id}/*",
+			"Condition": {
+				"StringEquals": {
+					"aws:SourceAccount": "${var.account_id}"
+				},
+				"ArnLike": {
+					"aws:SourceArn": "arn:aws:sns:${var.region}:${var.account_id}:*"
```
patheard, line 493: This is the spot that I think the issues lies:
```suggestion
					"aws:SourceArn": "arn:aws:sns:us-west-2:${var.account_id}:*"
```

aws/common/s3.tf
```diff
@@ -376,3 +376,159 @@ module "cbs_logs_bucket" {
...
+				"Service": "sns.amazonaws.com"
+			},
+			"Action": "s3:GetBucketLocation",
+			"Resource": "arn:aws:s3:::${module.sns_sms_usage_report_bucket_us_west_2.s3_bucket_id}",
+			"Condition": {
+				"StringEquals": {
+					"aws:SourceAccount": "${var.account_id}"
+				},
+				"ArnLike": {
+					"aws:SourceArn": "arn:aws:sns:${var.region}:${var.account_id}:*"
```
patheard, line 510: Same region change here:
```suggestion
					"aws:SourceArn": "arn:aws:sns:us-west-2:${var.account_id}:*"
```

aws/common/sns.tf
```diff
@@ -83,17 +83,26 @@ resource "aws_sns_topic" "notification-canada-ca-alert-general" {
 }
 
 resource "aws_sns_sms_preferences" "update-sms-prefs" {
+  depends_on = [
+    module.sns_sms_usage_report_bucket
+  ]
```
patheard, line 88: You can get rid of this `depends_on` block since this further down creates an implicit dependency:
```hcl
usage_report_s3_bucket                = module.sns_sms_usage_report_bucket.s3_bucket_id
```
```suggestion
```

aws/common/sns.tf
```diff
@@ -83,17 +83,26 @@ resource "aws_sns_topic" "notification-canada-ca-alert-general" {
...
   delivery_status_success_sampling_rate = 100
   monthly_spend_limit                   = var.sns_monthly_spend_limit
+  usage_report_s3_bucket                = module.sns_sms_usage_report_bucket.s3_bucket_id
 }
 
 resource "aws_sns_sms_preferences" "update-sms-prefs-us-west-2" {
+  depends_on = [
+    module.sns_sms_usage_report_bucket_us_west_2
+  ]
+
```
patheard, line 99: Same comment about this not being needed:
```suggestion
```

aws/common/s3.tf
jimleroyer: I can put in there the direct literal for the region, but I think `var.region` changes based on the selected provider:
https://github.com/cds-snc/notification-terraform/blob/c6e119e45498f5cc2da9d2c29516132649dec97c/env/terragrunt.hcl#L64

Does that make sense?

aws/common/sns.tf
jimleroyer: true

patheard (approved):

PR #1179, opened 2024-02-22
Refactor notifications dataset

sastels:
# Summary | Résumé

Uses SQL to create a new `notifications` dataset. This is a joined dataset that contains data from both the `notifications` and `notification_history` RDS tables.

We also remove the `notification_history` dataset since that dataset is a subset of the new `notifications` dataset.

Note that we can create this SQL dataset without needing to use CloudFormation, leading to much smaller and easier to understand / write code than what we had before 🎉 

I tested this by doing `terragrunt apply`'s on staging until I got the code working. Don't tell Pat. 😶


jimleroyer (approved):
LGTM! Let's try to see if that works out of the box!

PR #1190, opened 2024-03-06
Add latest tag to blazer image

jzbahrai:
# Summary | Résumé

We want to allow staging to use the latest blazer tag, and production will use the custom one


jimleroyer (approved):
Makes sense! 👍

PR #1215, opened 2024-03-21
Revert "Quicksight: smsusage leftjoin notifications"

jimleroyer:
# Description

This is a prevemptive reverts cds-snc/notification-terraform#1214, because QuickSight changes in TF tend to break. This is because of the immature documentation and plugins around Quicksight so while the plan might pass and we follow documentation, it tends to fail while the `terragrunt apply` step happens.


ben851 (approved):
Plan for failure!

PR #1217, opened 2024-03-21
Declare only 1 source for logical table source on smsusage dataset

jimleroyer:
# Summary | Résumé

On my last attempt to fix the TF changes to left join both smsusage and notifications dataset of quicksight, we got the following error:

```
Message_: "Exactly one of PhysicalTableId, JoinInstruction or DataSetArn must be set for LogicalTable"
```

Hence it seems that declaring the `PhysicalTableId` and a `JoinInstruction` are considered two different sources, even though the physical table is actually used in the join. I assume that the join instruction will get the references right. The documentation did not declare this constraint by the way, of having only one of these. Let's try this to see if that will fix our current issue and move on to the next one.

* https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/309

* Plan should pass.
* The terragrunt apply op should pass.

Monitor the `merge to main` github action closely to make sure it is successful. 

### DIFF
--- a/aws/quicksight/dataset_sms_usage.tf
+++ b/aws/quicksight/dataset_sms_usage.tf
@@ -54,7 +54,12 @@ resource "aws_quicksight_data_set" "sms_usage" {
     alias = "smsusage"
 
     source {
-      physical_table_id = "smsusage"
+      join_instruction {
+        left_operand  = "smsusage"
+        right_operand = aws_quicksight_data_set.notifications.data_set_id
+        on_clause     = "smsusage.MessageId = notifications.notification_id"
+        type          = "LEFT"
+      }
     }
 
     data_transforms {
--- a/aws/quicksight/dataset_sms_usage.tf
+++ b/aws/quicksight/dataset_sms_usage.tf
@@ -49,15 +49,16 @@ resource "aws_quicksight_data_set" "sms_usage" {
   }
 
   logical_table_map {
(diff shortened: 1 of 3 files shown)


aws/quicksight/dataset_sms_usage.tf
```diff
@@ -54,7 +54,12 @@ resource "aws_quicksight_data_set" "sms_usage" {
     alias = "smsusage"
 
     source {
-      physical_table_id = "smsusage"
```
jimleroyer, line 57: The error message that both physical table id and join instructions were present when only one should be. So we are removing that one!

ben851 (approved):
WORK!

aws/quicksight/dataset_sms_usage.tf
sastels: riiiight.... it wants just one think as the "source", ie here we want the source to be the join, basically.

I thought there also needed to be some code around what fields would appear in the final table... though maybe it's just going to use all of them if you don't restrict it? Best to apply on dev and see what happens 😿

sastels (approved):

jimleroyer:
Closing this as more recent changes went past the issue with the join by using CloudFormation instead of the Terraform QuickSight module.

PR #1219, opened 2024-03-22
Tweak QuickSight datasets

sastels:
# Summary | Résumé

Two minor tweaks:
- remove redundant code in `login_events` (we moved back to a full refresh and forgot to remove the incremental code)
- change field types in `notifications` from `BIT` (which seems to actually be cast to `INTEGER` in QuickSight) to the more correct `BOOLEAN`

in staging aft (or on dev if we get that sorted out)
- verify in staging that the `login_events` table still has a (full) refresh scheduled daily
- verify the types of the `service_active` and  `service_count_as_live` columns in `notifications`.

- notify users that the notification columns `service_active` and  `service_count_as_live` should be treated as boolean rather than integer. For example, filters should be on `service_active = true` rather than `service_active = 1`


P0NDER0SA (commented):
👀

aws/quicksight/dataset_notifications.tf
```diff
@@ -127,11 +127,11 @@ resource "aws_quicksight_data_set" "notifications" {
       }
       columns {
         name = "service_active"
-        type = "BIT"
+        type = "BOOLEAN"
       }
       columns {
         name = "service_count_as_live"
-        type = "BIT"
+        type = "BOOLEAN"
```
jimleroyer, line 134: Oh yes better; that might break some dashboards so we'll need to fix.

jimleroyer (approved):
LGTM!

sastels:
Followup 😞 
https://docs.aws.amazon.com/quicksight/latest/user/supported-data-types-and-values.html

I don't know where I got the list of available types but clearly it's wrong, any bit or boolean values get turned into integers by quicksight :/

PR #1246, opened 2024-04-10
Pinpoint pools

sastels:
# Summary | Résumé

- add Cloudwatch log groups for Pinpoint receipts (and role to allow Pinpoint to write to them)
- add repo to ECR for the [new Pinpoint receipts lambda function](https://github.com/cds-snc/notification-lambdas/pull/73).
- add lambda function for the [new Pinpoint receipts lambda function](https://github.com/cds-snc/notification-lambdas/pull/73) together with logs and and alarms
- add a script to create the Pinpoint Pools / configuration set. Unforutnately the Pinpoint SMS and Voice v2 API is not yet supported by terraform nor cloudformation.

Note that we make no changes to the existing SMS / SNS flow so this should have no impact on how the system runs (ie until the notify api starts directing texts to Pinpoint)

* https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/286

Have been building things on dev.

None - no affect on production. But at some point we should add these new alarms / metrics to the dashboards.


aws/pinpoint_to_sqs_sms_callbacks/lambda.tf
```diff
@@ -0,0 +1,65 @@
...
+  timeout                    = 60
+  memory                     = 1024
+  log_group_retention_period = var.sensitive_log_retention_period_days
+
+  policies = [
+    data.aws_iam_policy_document.pinpoint_to_sqs_sms_callbacks.json
+  ]
+}
+
+data "aws_sqs_queue" "delivery-receipts" {
```
ben851, line 17: Do we have this value somewhere else? I think in the sqs section? Would be good to have it referencing that in case we change it down the line.

ben851 (changes requested):
This is an amazing amount of work! Nice!

Can we move the cloudwatch logs, and iam from common into the pinpoint directory? That way everything related to pinpoint is in one spot (and you don't have to depend on common and outputs as a bonus!)

aws/common/cloudwatch_log.tf
```diff
@@ -44,6 +44,24 @@
...
+}
+
+# Pinpoint will log delivery receipts here, so it needs to be created
+resource "aws_cloudwatch_log_group" "pinpoint_deliveries_failures" {
+  name              = "sns/${var.region}/${var.account_id}/PinPointDirectPublishToPhoneNumber/Failure"
+  retention_in_days = var.sensitive_log_retention_period_days
+  tags = {
+    CostCenter = "notification-canada-ca-${var.env}"
+  }
+}
```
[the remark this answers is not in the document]
ben851: We don't do this yet, but we should add a card to get this done on all of the logs.

aws/pinpoint_to_sqs_sms_callbacks/lambda.tf
sastels: yesss we can pull it from common 👍

aws/ecr/images.tf
```diff
@@ -193,7 +193,7 @@ resource "null_resource" "build_sns_to_sqs_sms_callbacks_docker_image" {
   ]
 
   provisioner "local-exec" {
-    command = "docker build -t ${aws_ecr_repository.sns_to_sqs_sms_callbacks.repository_url}:bootstrap -f /var/tmp/notification-lambdas/sesemailcallbacks/Dockerfile /var/tmp/notification-lambdas"
```
sastels, line 196: Noticed this typo

ben851 (changes requested):
It's a cool idea to keep this contained with the lambda function, but it won't work as is because on a new environment apply, the lambda module will fail because there is no dependency relationship between the module and these (so the module will try and run before the docker image exists).

If you add a depends_on to the module, it won't work either because the dependency resource only creates when bootstrap is set to true.

You could try a conditional dependency - I'm not sure that's a thing... but something like:

depends_on = [var.bootstrap ? null_resource.push_pinpoint_to_sqs_sms_callbacks_docker_image : null ]

no idea if that will work though... If it doesn't, it's probably easiest to just move this file back to the ecr directory.

Also, does the create_pinpoint_pools.sh script have to be run before deploy? If that's the case, can we add it as a null_resource that gets executed if bootstrap = true?

sastels:
> Also, does the create_pinpoint_pools.sh script have to be run before deploy? If that's the case, can we add it as a null_resource that gets executed if bootstrap = true?

no, in fact we need to have the terraform done (at least we need the iam role and cloudwatch receipt log groups).

ben851 (approved):
LGTM!

PR #1248, opened 2024-04-11
fix: RDS activity stream deterministic zip

patheard:
# Summary
Upgrade the module to so that the decrypt Lambda dependency layer is only recreated when the dependencies change.

# Related
- https://github.com/cds-snc/platform-core-services/issues/508


ben851 (approved):
LGTM

PR #1275, opened 2024-04-23
DNS Terraform Updates

P0NDER0SA:
# Summary | Résumé

condensing multiple files that used to create these resources to include the environment names, to not use the environment names, just like everything else.  These should only be run when not production.

This was something discsovered during BCP work.

https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/58

Watch the pipeline to see how the TF workflows react.


P0NDER0SA:
I"m seeing that there are discrepencies between the old staging and dev files... need to discuss why there are 10 resources in one and only 7 in the other

ben851 (approved):
I think this LGTM

PR #1306, opened 2024-05-07
Removing the duplication of code

ben851:
# Summary | Résumé

We had duplicate roles for logging on the api gateway. For some reason the original wasn't working, but now that we can see that it was duplicated, we should focus on fixing the first and not having two.

Was originally part of BCP

This may require two runs of TF apply - one to remove the duplicate and one to set the correct original (they were squashing each other so it depends on the run order)

None.


P0NDER0SA (approved):
OK, this makes sense as it already existed.

PR #1367, opened 2024-06-10
Revert "restoring new static site"

ben851:
Reverts cds-snc/notification-terraform#1361

Reverting back to old version since we don't want this to go to prod at the moment.


sastels (approved):
Revert!!

PR #1411, opened 2024-07-03
reverting oidc

ben851:
# Summary | Résumé

Reverting OIDC to see if this was the problem with TF plan/apply


sastels (approved):
ok let's try...

PR #1420, opened 2024-07-08
Adding git checkout to bump version

ben851:
# Summary | Résumé

Bump version is not working, we think it's because git checkout is not run.


sastels (approved):
Giv'r!

PR #1484, opened 2024-08-01
Revert "remove send_rate dataset for now"

sastels:
# Summary | Résumé

Reverts cds-snc/notification-terraform#1403
This should work now that we've increased the instance class.

Add a new dataset so we can have a send rate dashboard in QuickSight.

~**NOTE**: releasing this to production will probably trigger the large refresh job for this dataset. Our system will be ok with this, but we probably don't want to do it at the same time #1478 releases (otherwise we might have 2 of our instances busy for an hour?). So let's not merge this PR until after #1478 is released.~

Verify dataset on staging

Verify dataset on production.

- [x] This PR does not break existing functionality.
- [x] This PR does not violate GCNotify's privacy policies.
- [x] This PR does not raise new security concerns. Refer to our GC Notify Risk Register document on our Google drive.
- [x] This PR does not significantly alter performance.
- [x] Additional required documentation resulting of these changes is covered (such as the README, setup instructions, a related ADR or the technical documentation).


ben851 (approved):
What's the worst that could happen?

jimleroyer:
We should also take a look at doing join tables using the Quisight plugin, now that the bug for it has been resolved. This could open up the door for us to do refresh instead of complete updates and be more efficient, if I understand the issues and chain of events correctly.

sastels:
> We should also take a look at doing join tables using the Quisight plugin, now that the bug for it has been resolved. This could open up the door for us to do refresh instead of complete updates and be more efficient, if I understand the issues and chain of events correctly.

I don't think that would work here, where we're doing a double groupby, but definitely we should be able to change the way we do the notifications dataset.

PR #1526, opened 2024-09-10
Update modules (the revert of the revert)

jzbahrai:
We were failing during the github action, Ben ran the action locally and got it to pass. We are now going to do the upgrade for the relevant TF modules


ben851 (approved):
Should probably wait till after the release this morning before merging but LGTM

PR #1763, opened 2025-02-13
Bring back limited send_rate dataset

sastels:
Reverts cds-snc/notification-terraform#1403


sastels:
This was basically for the perf test. Instead we added a plql query directly to the perf test and save / post the results.

PR #1849, opened 2025-04-15
Create new SQS buffer queue for ses receipt callbacks

whabanks:
# Summary | Résumé

This PR implements the infrastructure changes outlined in the [batch saving receipts ADR](https://github.com/cds-snc/notification-adr/pull/81) to add a new buffer SQS queue that will enable the `ses_to_sqs_email_callbacks` lambda to receive receipts in batches of 10

### Before
The `ses_to_sqs_email_callbacks` lambda was subscribed directly to the SNS topic that AWS SES publishes email receipts to. This limited the lambda to receiving a single receipt at a time as SNS topics do not allow subscribers to pull more than one message at a time from the topic.

image

### After
The new SQS queue is added and subscribed to the aforementioned SNS topic, and the aforementioned lambda is now subscribed to the SQS queue, allowing it to pull SES receipts in batches.

image

### Changes
- Added SQS queue `ses-receipt-callback-buffer`
- Granted ses_to_sqs_email_callbacks rights to the new SQS queue: `sqs:GetQueueAttributes`, `sqs:ReceiveMessage`, and `sqs:DeleteMessage`
- Subscribed `ses-receipt-callback-buffer` to the `notification-canada-ca-ses-callback` SNS topic
- Unsubscribed `ses_to_sqs_email_callbacks` lambda from the above SNS topic
- Subscribed the lambda to the new SQS queue
- Added a new IAM policy to ensure that only the new SQS queue can receive messages from the aforementioned SNS topic 

## Related Issues | Cartes liées

* https://github.com/cds-snc/notification-adr/commit/a794783e89bd85a6c0bb37aa4c346a5a0ca6c813
* https://github.com/cds-snc/notification-lambdas/pull/96
* https://github.com/cds-snc/notification-api/pull/2489

## Before merging this PR

Read code suggestions left by the
[cds-ai-codereviewer](https://github.com/cds-snc/cds-ai-codereviewer/) bot. Address
valid suggestions and shortly write down reasons to not address others. To help
with the classification of the comments, please use these reactions on each of the
comments made by the AI review:

| Classification      | Reaction | Emoticon |
|---------------------|----------|----------|
| Useful              | +1       | 👍        |
| Noisy               | eyes     | 👀        |
| Hallucination       | confused | 😕        |
| Wrong but teachable | rocket   | 🚀        |
| Wrong and incorrect | -1       | 👎        |

The classifications will be extracted and summarized into an analysis of how helpful
or not the AI code review really is.

[...]

PR #1870, opened 2025-05-01
Revert "Create new SQS buffer queue for ses receipt callbacks"

whabanks:
A "just in case" revert PR for the batch saving infra if merge to staging doesn't go well: cds-snc/notification-terraform#1849


ben851 (approved):
LGTM

PR #2004, opened 2025-06-24
Revert "force VPN cert recreate"

ben851:
Reverts cds-snc/notification-terraform#2003


jimleroyer (approved):
That didn't work out?

PR #2020, opened 2025-07-02
Elasticache Split Again

ben851:
# Summary | Résumé

Splitting elasticache back into two so I can debug this in staging.

## Related Issues | Cartes liées

*  https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/565

## Test instructions | Instructions pour tester la modification

TF Apply
Smoke Tests
Debug!

## Release Instructions | Instructions pour le déploiement

None.

## Reviewer checklist | Liste de vérification du réviseur

* [ ] This PR does not break existing functionality.
* [ ] This PR does not violate GCNotify's privacy policies.
* [ ] This PR does not raise new security concerns. Refer to our GC Notify Risk Register document on our Google drive.
* [ ] This PR does not significantly alter performance.
* [ ] Additional required documentation resulting of these changes is covered (such as the README, setup instructions, a related ADR or the technical documentation).

> ⚠ If boxes cannot be checked off before merging the PR, they should be moved to the "Release Instructions" section with appropriate steps required to verify before release. For example, changes to celery code may require tests on staging to verify that performance has not been affected.

PR #2021, opened 2025-07-02
Revert "Elasticache Split Again"

ben851:
Reverts cds-snc/notification-terraform#2020

I can at least replicate the issue on dev now so I should be able to put staging back.


P0NDER0SA (approved):
Looks good

PR #2459, opened 2026-04-08
set a statistically significant data point

jzbahrai:
# Summary | Résumé

Change metric to be statistically significant


P0NDER0SA (approved):
Yup, this should help us in the low send time frames like long weekends

PR #2593, opened 2026-05-28
Update waf.tf to include v3 endpoints

jzbahrai:
# Summary | Résumé

We are adding v3 endpoints to waf as the new template endpoints will be behind the v3 flag (they new enpoints require a manage_template permission for the API key)


P0NDER0SA (approved):
yep, looks ok

PR #2618, opened 2026-06-10
Fix the expired inflight alarm

smcmurtry:
# Summary | Résumé

The `expired-inflight-critical` and `expired-inflight-warning` alarms were effectively always firing during high-volume bulk sends because the `inflight_processed` metric query used `{acknowledged: True}` as its only dimension. CloudWatch requires exact dimension matching, but the API never emitted a metric with that dimension set alone - so the query always returned no data, making the "zero throughput" condition permanently true.

Changes:

- Removed the two aggregate alarms and replaced them with 12 per-queue alarms (6 critical, 6 warning) covering each combination of: `email/sms × bulk/normal/priority`
- Per-queue alarms use specific dimensions that the API already emits, so no API changes are needed
- Updated the EKS alarms dashboard to reference the new per-queue critical alarm names

This also means the alarms now correctly detect when a single queue is stuck while others are healthy - which the aggregate alarms would have missed previously.

Not sure if it is possible to test this.

None.


aws/common/cloudwatch_alarms.tf
```diff
@@ -979,7 +979,9 @@ resource "aws_cloudwatch_metric_alarm" "expired-inflight-critical" {
       stat        = "Sum"
       unit        = "Count"
       dimensions = {
-        acknowledged = "True"
+        acknowledged      = "True"
+        notification_type = "any"
+        priority          = "any"
```
smcmurtry: @jimleroyer / @ben851 I'm not sure if this is right or not - do we want to narrow this down so that we know if a particular queue is stuck?

jzbahrai (approved):
good design change, this makes sense to me

ben851 (approved):
LGTM

aws/common/cloudwatch_alarms.tf
```diff
@@ -979,7 +979,9 @@ resource "aws_cloudwatch_metric_alarm" "expired-inflight-critical" {
...
       }
```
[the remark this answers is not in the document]
smcmurtry: fixed

aws/common/cloudwatch_alarms.tf
smcmurtry: decided to go with separate alarms

aws/common/cloudwatch_alarms.tf
```diff
@@ -964,11 +967,23 @@ resource "aws_cloudwatch_metric_alarm" "expired-inflight-warning" {
...
-  alarm_name          = "expired-inflight-critical"
-  alarm_description   = "More than ${var.alarm_critical_expired_inflights_threshold} inflights expired in 5 minutes AND celery acknowledgment throughput is zero - system is stuck, check the Redis-batch-saving dashboard"
+locals {
+  inflight_queues = [
+    { name = "email-bulk", notification_type = "email", priority = "bulk" },
+    { name = "email-normal", notification_type = "email", priority = "normal" },
+    { name = "email-priority", notification_type = "email", priority = "priority" },
+    { name = "sms-bulk", notification_type = "sms", priority = "bulk" },
+    { name = "sms-normal", notification_type = "sms", priority = "normal" },
+    { name = "sms-priority", notification_type = "sms", priority = "priority" },
```
jimleroyer, line 977: ah that's nice

PR #2677, opened 2026-07-13
Update duplicate-record alarm to use structured log tag; add unexpected-constraint alarm

smcmurtry:
## What

Updates the `celery-error-duplicate-record` CloudWatch log filter and adds a new `celery-error-unexpected-db-constraint` filter + alarm pair.

## Why

The notification-api change ([cds-snc/notification-api#2947](https://github.com/cds-snc/notification-api/pull/2947)) classifies expected duplicate-record errors from SQS redelivery with the structured tag `CELERY_KNOWN_ERROR::DUPLICATE_RECORD`. This allows the two cases to be monitored separately:

| Filter | Matches | Meaning |
|---|---|---|
| `celery-error-duplicate-record` | `CELERY_KNOWN_ERROR::DUPLICATE_RECORD` | Known-safe dedup — SQS redelivery caught and handled correctly |
| `celery-error-unexpected-db-constraint` | raw Postgres string **without** the structured tag | Unexpected constraint violation in a different code path — investigate |

Previously both cases matched the same raw Postgres string filter, so the alarm was firing on expected behaviour introduced by cds-snc/notification-api#2944.

## Dependency

Requires cds-snc/notification-api#2947 to be deployed before this terraform change takes effect. Until then the `celery-error-duplicate-record` alarm will stop firing (because the structured tag won't appear in logs yet) — this is acceptable since the current alarm is firing on expected behaviour anyway.

### DIFF
--- a/aws/eks/cloudwatch_alarms.tf
+++ b/aws/eks/cloudwatch_alarms.tf
@@ -193,6 +193,42 @@ resource "aws_cloudwatch_metric_alarm" "logs-celery-error-duplicate-record-criti
   ok_actions          = [var.sns_alert_ok_arn]
 }
 
+# UNEXPECTED_DB_CONSTRAINT — raw unique-constraint violations not caught by the
+# known-safe dedup flow. Any hit here is unexpected and should be investigated.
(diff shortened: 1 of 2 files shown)


smcmurtry:
Going with a simpler approach

PR #2724, opened 2026-07-29
waf refinements and queries

ben851:
# Summary | Résumé

Adding more tuning to the waf.

There are a bunch of rules in count that will show false positives due to how the waf rules behave with labels. I've added athena queries that will filter out false positives that we can use in the future to determine if the waf is working as expected.

*  https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/875

TF Apply
Watch prod after release

None.

### DIFF
--- a/aws/eks/waf.tf
+++ b/aws/eks/waf.tf
@@ -29,7 +29,7 @@
 #  20 |   200 | AWSManagedRulesLinuxRuleSet                  | BLOCK (managed)
 #  21 |   700 | AWSManagedRulesCommonRuleSet                 | BLOCK (managed, sets labels)
 #  22 |    ~8 | BlockLabeled_SSRF_NoUserAgent_NonCA          | count → BLOCK (label, after 21, non-API)
-#  23 |    ~6 | BlockSizeRestrictions_Body_ExcludeUploadPaths| count → BLOCK (label, after 21)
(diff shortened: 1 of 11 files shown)


aws/eks/waf.tf
```diff
@@ -1508,6 +1511,26 @@ resource "aws_wafv2_web_acl" "notification-canada-ca" {
             key   = "awswaf:managed:aws:core-rule-set:SizeRestrictions_BODY"
           }
         }
+        # Exclude API host: body size on api is enforced by BlockLargeRequests_Body_Api (7 MB)
```
jimleroyer, line 1514: Related but not applicable here: is the 7MB large request for API inherited from lambda's limitations? I know we could handle 10 or 12 MB when the API was in Kubernetes, and when we moved the API to the lambda, we had to lower that limit down.

jimleroyer (approved):
Very nice addition of the queries. This is an area we need to improve and this is great to see these.

aws/eks/waf.tf
ben851: I just picked a number honestly - I know we had a 6mb limit on file attachments in the admin so I made it slightly larger than that. I'm happy to change this number to anything that makes more sense

PR #2736, opened 2026-08-14
Check if a service is sending more than 100 international messages per hour

jzbahrai:
# Summary
Adds alerting when any single service sends a high volume of international SMS messages, so we can proactively control costs.

## How it works
1. Emits a CloudWatch EMF metric (international_sms_sent) with service_id as a dimension each time an international SMS is sent
1. A Metrics Insights alarm with GROUP BY service_id evaluates all services independently — no service list to maintain
1. The alarm fires when any individual service sends ≥100 international SMS in a 1-hour window
1. A Logs Insights saved query is available for triage to see which services are sending international messages

## Changes
1. cloudwatch_alarms.tf — Metrics Insights query alarm with GROUP BY service_id
1. cloudwatch_queries.tf — Saved query "SMS / International SMS by service" for triage

## Threshold
The threshold of 100/hour is a starting point.

## Related PR:
https://github.com/cds-snc/notification-api/pull/3008


whabanks (approved):
LGTM!

PR #2739, opened 2026-08-18
remove api large request waf rule

ben851:
# Summary | Résumé

Removing the API WAF rule for governing the maximum request size. It will not work on the WAF:

After reviewing, we can't block large requests for the API at the WAF level. AWS WAF only reads the first 8kb of the body, and anything after that would automatically trip it to "too large" and block the request. Thus there's no way to differentiate between 9kb and 7MB. They suggest doing this at the ALB level, but we can't do that either because we don't use nginx (we use AWS ALB Controller). The last suggestion was to do it at the flask level, but I'm not sure if that's worth it.

* https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/875

TF Apply

None.

### DIFF
--- a/aws/common/athena_queries_support.tf
+++ b/aws/common/athena_queries_support.tf
@@ -138,7 +138,7 @@ resource "aws_athena_named_query" "waf_admin_geo_restriction_hits" {
 
 resource "aws_athena_named_query" "waf_size_restrictions_body_hits" {
   name        = "WAF: SizeRestrictions body hits"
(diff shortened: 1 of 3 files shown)


jimleroyer (approved):
Your PR description rings a bell. We looked into this years ago and came to the same conclusion. One thing: We do restrict the max request size, but for attachments. So we have something similar at the app level in Flask, but not exactly it. It feels late though at the app level to check the overall request size.

PR #2755, opened 2026-08-21
Add checks for blazer

jzbahrai:
## Summary
This PR enables automated Blazer check execution and wires Blazer check notifications to Slack for the database-tools deployment.

## Why
Blazer had check schedules configured in app settings, but infrastructure did not run the checks on a schedule and did not pass a Slack webhook into the running task. This meant checks were not consistently executed and Slack notifications were not fully wired.

## New user as well
This Terraform PR introduces a separate connection path for Blazer checks so checks can run with a restricted database user while normal Blazer querying remains unchanged: it adds a new sensitive input variable and SecureString parameter for blazer_checks_database_url in variables.tf and parameterstore.tf, injects that value into the Blazer ECS task as BLAZER_CHECKS_DATABASE_URL in blazer.tf, and updates both execution and task IAM policies to allow reading the new parameter in container_execution_role.tf and task_role.tf

## How eventbridge reaches out to blazer
1. Scheduled rules are created in check_schedules.tf:18 with cron/rate expressions at check_schedules.tf:22.
1 .Each rule targets the Blazer ECS cluster ARN, not an HTTP endpoint, at check_schedules.tf:30 and check_schedules.tf:36.
1. EventBridge assumes a role at check_schedules.tf:37, where trust allows events.amazonaws.com in check_schedules_iam.tf:6, and permissions allow ecs:RunTask and iam:PassRole at check_schedules_iam.tf:15 and check_schedules_iam.tf:22.
1. EventBridge starts a one-off Fargate task from the Blazer task definition at check_schedules.tf:58, overriding the container command to run blazer:run_checks at check_schedules.tf:40.
1. That task runs inside private subnets with no public IP at check_schedules.tf:62, using the same VPC network model as the Blazer service at blazer.tf:22 and blazer.tf:24.

## What changed
1. Inject Blazer Slack webhook into the ECS task definition:
2. Create a secure SSM parameter for the Blazer Slack webhook:
3. Allow Blazer task and execution roles to read the new webhook parameter:
4. Add scheduled EventBridge rules to run Blazer checks as one-off ECS tasks:
- every 5 minutes with SCHEDULE=5 minutes
- every 1 hour with SCHEDULE=1 hour
- daily at 9am est/10am edt with SCHEDULE=1 day
`check_schedules.tf`

5. Add IAM role and policy for EventBridge to run ECS tasks and pass required roles:
`check_schedules_iam.tf`

Behavior notes
- Schedules are only created when cloudwatch_enabled is true.
- Scheduled runs use the existing Blazer ECS task definition and private subnet networking.

Testing

1. Trigger one rule manually and confirm:
- ECS task starts successfully
- Blazer checks execute
- Slack notification is received for failing/passing checks as configured

Risk and rollback
- Risk is low and scoped to database-tools Blazer scheduling and secret wiring.
- Rollback is straightforward by reverting this PR, which removes schedules and Slack webhook injection.

### DIFF
--- a/aws/database-tools/blazer.tf
+++ b/aws/database-tools/blazer.tf
@@ -68,6 +68,9 @@ resource "aws_ecs_task_definition" "blazer" {
         "name" : "BLAZER_DATABASE_URL",
         "valueFrom" : "${aws_ssm_parameter.sqlalchemy_database_reader_uri.arn}"
         }, {
+        "name" : "BLAZER_CHECKS_DATABASE_URL",
+        "valueFrom" : "${aws_ssm_parameter.blazer_checks_database_url.arn}"
+        }, {
         "name" : "DATABASE_URL",
         "valueFrom" : "${aws_ssm_parameter.db_tools_environment_variables.arn}"
         }, {
--- a/aws/database-tools/blazer.tf
+++ b/aws/database-tools/blazer.tf
@@ -71,6 +71,9 @@ resource "aws_ecs_task_definition" "blazer" {
         "name" : "DATABASE_URL",
         "valueFrom" : "${aws_ssm_parameter.db_tools_environment_variables.arn}"
         }, {
+        "name" : "BLAZER_SLACK_WEBHOOK_URL",
+        "valueFrom" : "${aws_ssm_parameter.blazer_slack_webhook_general_topic.arn}"
+        }, {
         "name" : "GOOGLE_OAUTH_CLIENT_ID",
         "valueFrom" : "${aws_ssm_parameter.notify_o11y_google_oauth_client_id.arn}"
         }, {
--- a/aws/database-tools/check_schedules_iam.tf
+++ b/aws/database-tools/check_schedules_iam.tf
@@ -1,0 +1,45 @@
+data "aws_iam_policy_document" "scheduled_task_blazer_event_role_assume_role_policy" {
+  statement {
+    effect  = "Allow"
+    actions = ["sts:AssumeRole"]
+    principals {
+      identifiers = ["events.amazonaws.com"]
+      type        = "Service"
+    }
+  }
+}
+
+data "aws_iam_policy_document" "scheduled_task_blazer_event_role_cloudwatch_policy" {
+  statement {
+    effect  = "Allow"
+    actions = ["ecs:RunTask"]
+    #checkov:skip=CKV_AWS_111:Required to run scheduled Blazer ECS tasks
+    resources = [aws_ecs_task_definition.blazer.arn]
+  }
+
+  statement {
+    effect  = "Allow"
+    actions = ["iam:PassRole"]
+    resources = [
+      aws_iam_role.blazer_ecs_task.arn,
+      aws_iam_role.blazer_execution_role.arn
+    ]
+  }
+}
+
+resource "aws_iam_role" "scheduled_task_blazer_event_role" {
+  provider           = aws.core_services
+  name               = "blazer-scheduled-task-role-${var.env}"
+  assume_role_policy = data.aws_iam_policy_document.scheduled_task_blazer_event_role_assume_role_policy.json
+
+  tags = {
+    (var.billing_tag_key) = var.billing_tag_value
+  }
+}
+
+resource "aws_iam_role_policy" "scheduled_task_blazer_event_role_cloudwatch_policy" {
+  provider = aws.core_services
+  name     = "${aws_ecs_cluster.blazer.name}-ecs-scheduled-policy-${var.env}"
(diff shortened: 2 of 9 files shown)


aws/database-tools/check_schedules_iam.tf
```diff
@@ -0,0 +1,45 @@
...
+
+  statement {
+    effect  = "Allow"
+    actions = ["iam:PassRole"]
+    resources = [
+      aws_iam_role.blazer_ecs_task.arn,
+      aws_iam_role.blazer_execution_role.arn
+    ]
+  }
+}
```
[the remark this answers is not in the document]
jzbahrai: fixed

aws/database-tools/blazer.tf
```diff
@@ -68,9 +68,15 @@ resource "aws_ecs_task_definition" "blazer" {
         "name" : "BLAZER_DATABASE_URL",
         "valueFrom" : "${aws_ssm_parameter.sqlalchemy_database_reader_uri.arn}"
         }, {
+        "name" : "BLAZER_CHECKS_DATABASE_URL",
```
ben851, line 71: Is this different from the regular postgres URL defined above?

aws/database-tools/parameterstore.tf
```diff
@@ -38,6 +38,30 @@ resource "aws_ssm_parameter" "notify_o11y_google_oauth_client_secret" {
   }
 }
 
+resource "aws_ssm_parameter" "blazer_slack_webhook_general_topic" {
+  provider = aws.core_services
+  name     = "blazer_slack_webhook_general_topic"
+  type     = "SecureString"
+  value    = var.blazer_slack_webhook_general_topic
```
ben851, line 45: I think this is just the regular slack webhook for general which is already defined in the secrets? If so you can just use this
```suggestion
  value    = var. cloudwatch_slack_webhook_general_topic
```

aws/database-tools/parameterstore.tf
```diff
@@ -38,6 +38,30 @@ resource "aws_ssm_parameter" "notify_o11y_google_oauth_client_secret" {
...
+    (var.billing_tag_key) = var.billing_tag_value
+    Terraform             = true
+  }
+}
+
+resource "aws_ssm_parameter" "blazer_checks_database_url" {
+  provider = aws.core_services
+  name     = "blazer_checks_database_url"
+  type     = "SecureString"
+  value    = var.blazer_checks_database_url
```
ben851: The database URLs are all programmatically handled in Terraform, we should reference it via a dependency from the rds module... Again, not sure if this is different than the regular postgres connection url?

ben851 (changes requested):

aws/database-tools/blazer.tf
jzbahrai: I added this checks database that also queries against the prd DB, but it is its own datastore. This way the checks user will only run on the checks database (defined in the above url). We can restrict access to tables in this datastore. This is a blaze configuration that will allow us to do so.

Here's the AI defn if that helps: 
Blazer lets you configure multiple datasources by name . The intent here (per the PR's own description, "New user as well") is that unattended, scheduled checks run under a different, more restricted DB role than a logged-in human using the interactive UI — so if a check query is compromised, mis-scheduled, or buggy, it's limited to whatever tables that restricted user can see, while a human's interactive session keeps its existing (broader) app_db_user access via BLAZER_DATABASE_URL. It's a least-privilege split for the automation path, not a duplicate of the existing reader URL.


Env var | SSM source | Connects as | Purpose
-- | -- | -- | --
DATABASE_URL | db_tools_environment_variables (named "BLAZER_DATABASE_URL" in SSM, confusingly) | postgres superuser, to the dedicated database-tools RDS instance | Blazer's own internal app database — stores Blazer's saved queries, dashboards, users, check definitions/schedules, run history, etc. This is Blazer's Rails backing store, not the data being analyzed.
BLAZER_DATABASE_URL | sqlalchemy_database_reader_uri | app_db_user via the RDS read-only proxy | The default datasource Blazer queries against — i.e., the actual GC Notify production data that a human browses/queries ad hoc in the Blazer UI.
BLAZER_CHECKS_DATABASE_URL (new) | blazer_checks_database_url | a new, separate restricted user (not yet created) via the same read-only proxy | A second, named datasource Blazer's gem supports specifically for the automated blazer:run_checks scheduled task this PR adds.

aws/database-tools/parameterstore.tf
jzbahrai: done

aws/database-tools/parameterstore.tf
jzbahrai: same as above

ben851 (approved):
LGTM

PR #2763, opened 2026-08-24
switch of callback paging

jzbahrai:
# Summary | Résumé

Keeping the alarm for visibility. We switch of callbacks and email clients automatically.


ben851 (approved):
LGTM

PR #2787, opened 2026-09-09
Remove manifest_ prefix from Manifest repository secret variables and secret names

P0NDER0SA:
## What
Part 1: renames all `manifest_`-prefixed Terraform **variables** to drop the prefix (`env/variables.tf` etc.).

Part 2: adds de-prefixed **duplicate** AWS Secrets Manager secrets for all 44 manifest secrets in `aws/manifest_secrets/secrets.tf`, additive-only. The original `MANIFEST_*` secrets are untouched.

## Why additive, not a rename
Renaming the `name` attribute on `aws_secretsmanager_secret` is a ForceNew change — Terraform destroys and recreates the resource. These secrets are configured with `recovery_window_in_days = 0`, so an in-place rename would delete the old secret **immediately and irrecoverably**, before the Helm/ExternalSecret side (notification-manifests) is ready to read the new names.

Instead, each of the 44 manifest secrets now has a `"_unprefixed"` twin resource, sourcing from the same variable/expression, with the `MANIFEST_` prefix stripped from the AWS-facing name. Nothing is deleted by this PR.

## Rollout plan
1. Merge/apply this PR — creates 44 new secrets side by side with the existing ones. Zero risk, fully reversible.
2. Update the notification-manifests Helm/ExternalSecret config to read the new de-prefixed names, deploy, and verify in each environment.
3. Open a follow-up PR here to delete the legacy `MANIFEST_`-prefixed resources (their labels and the "LEGACY" comment block make them easy to find).

## Variable rename collisions handled
Three of the `manifest_` variables collided with existing declarations once the prefix was stripped:
- `manifest_admin_client_secret` → merged into the existing `admin_client_secret` variable (had zero consumers, no functional change).
- `manifest_auth_tokens` → merged into the existing `auth_tokens` variable (had zero consumers).
- `manifest_waf_secret` → merged into the existing `waf_secret` variable (confirmed same value used for both, consolidated into one).
- `document_download_api_key` (formerly `manifest_document_download_api_key`) is removed entirely; the secret now reads from `auth_tokens` since both held the same value.

## Not in scope / needs manual follow-up
- `.tfvars` files (`production.tfvars`, `production-keys.tfvars`, `staging.tfvars`, etc.) were updated separately, outside this PR.
- Resource local names (Terraform addresses) for the legacy resources are left unchanged.
- Corresponding 1Password entries should be double-checked against the new variable names.
- Downstream repos/services (notification-manifests Helm/ExternalSecret config) still need to be updated to read the new de-prefixed secret names — tracked separately.

### DIFF
--- a/aws/common/cloudwatch_events.tf
+++ b/aws/common/cloudwatch_events.tf
@@ -58,7 +58,7 @@ resource "aws_cloudwatch_event_connection" "guardduty_scan_verdict_callback" {
   auth_parameters {
     api_key {
       key   = "X-Scan-Callback-Token"
-      value = var.manifest_scan_verdict_callback_token
+      value = var.scan_verdict_callback_token
     }
   }
 }
--- a/aws/github/api-secrets.tf
+++ b/aws/github/api-secrets.tf
@@ -9,7 +9,7 @@ resource "github_actions_secret" "api_cypress_user_pw_secret" {
   count            = var.env == "staging" ? 1 : 0
   repository       = data.github_repository.notification_api.name
   secret_name      = "CYPRESS_USER_PW_SECRET"
-  plaintext_value  = var.manifest_cypress_user_pw_secret
+  plaintext_value  = var.cypress_user_pw_secret
   destroy_on_drift = false
 }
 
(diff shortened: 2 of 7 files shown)


ben851 (approved):
LGTM

Issue #78, opened 2020-11-20
Create S3 bucket for Health Canada


AntoineAugusti:

AntoineAugusti:
Done in https://github.com/cds-snc/notification-terraform/pull/79

PR #2798, opened 2026-09-15
Migrate Terraform image consumers to SHA tags

P0NDER0SA:
## Summary
Migrate Terraform image consumers away from mutable `latest` and `bootstrap` tags by reading authoritative image tags from `notification-manifests` and validating those tags in the correct regional ECR repositories.

## Details
- Reads `helmfile/overrides/${var.env}.env` from `notification-manifests` using the GitHub provider.
- Parses the relevant `*_DOCKER_TAG` value for every Terraform image consumer, including Blazer and both Pinpoint regions.
- Uses `aws_ecr_image` data sources so Terraform validates that the referenced tag exists before using it.
- Removes hardcoded Lambda/Blazer Docker tag assignments from normal Terraform consumer configuration.
- Preserves Lambda infrastructure, event wiring, regions, and aliases.

## Dependencies
- Producer and manifests flows are proven for the current SHA migration.
- Merge [notification-manifests #5252](https://github.com/cds-snc/notification-manifests/pull/5252) before applying production, because it seeds the two new production Lambda keys with verified SHA `283323b`.

## Phase 2
Terraform bootstrap image production and mutable `latest` cleanup remain a separate follow-up after this consumer migration is proven.

## Validation
- `git diff --check` passed.
- Terraform formatting checks passed.


P0NDER0SA:
Updated this PR onto current main. Production manifests still had `DOCKER_TAG` placeholders for the two migrated Lambdas, so notification-manifests PR #5251 seeds them with the verified production SHA `283323b`. Merge #5251 before applying the production Terraform consumer migration.

P0NDER0SA:
Correction: the production manifest dependency is [notification-manifests #5252](https://github.com/cds-snc/notification-manifests/pull/5252), not #5251.

P0NDER0SA:
Production promotion dependency clarified: manifests #5252 seeds the current verified SHA, and notification-pr-bot #78 adds SES/SNS to future automatic production release promotion. Terraform production consumption should wait for both.

PR #2799, opened 2026-09-16
Change the critical alarm to paging for international message sending

jzbahrai:
Change the critical alarm to paging for international message sending


ben851 (approved):
lgtm

PR #2802, opened 2026-09-17
Add least-privilege staging developer IAM identity

P0NDER0SA:
## Summary

- Add a staging-only Terraform-managed IAM user for local API, admin, and document-download development.
- Scope SES, SNS, Pinpoint/SMS Voice, S3, SQS, KMS, Security Hub, and STS permissions to staging resources where AWS supports resource scoping.
- Store the generated access key in staging Secrets Manager.
- Exclude production, EKS, IAM administration, and CloudWatch access.

## QA

- `terraform fmt -check` passed.
- Repository Terraform formatting pre-commit checks passed.
- Terraform plan/validation requires the staging AWS credentials and should run in CI/PR.
- Developer verification is required for allowed staging operations and denied production access.


aws/common/staging_developer_iam.tf
```diff
@@ -0,0 +1,200 @@
...
+    actions = [
+      "ses:GetIdentityVerificationAttributes",
+      "ses:ListIdentities",
+      "ses:SendEmail",
+      "ses:SendRawEmail",
+      "ses:VerifyEmailIdentity",
+    ]
+
+    resources = ["*"]
+  }
```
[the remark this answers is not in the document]
ben851: This is fine.

aws/common/staging_developer_iam.tf
```diff
@@ -0,0 +1,200 @@
...
+  provider = aws.core_services
+  user     = aws_iam_user.staging_developer[0].name
+}
+
+resource "aws_secretsmanager_secret" "staging_developer_credentials" {
+  count = local.staging_developer_enabled ? 1 : 0
+
+  provider                = aws.core_services
+  name                    = "staging/developer/aws-credentials"
+  recovery_window_in_days = 7
```
ben851:
```suggestion
  recovery_window_in_days = 0
```

ben851: If we have to delete and recreate the secret, this recovery window makes us wait a week. If TF was our single source of truth for things this would be fine since we'd want the recoverability, but since we use 1Pass for secrets, having this set to 0 makes our lives easier in the future. Also since this one is generated by code, if we want to delete/recreate the user, this would fail within that 7 day window.

ben851 (approved):
LGTM but made a suggestion

PR #2805, opened 2026-09-17
Add missing SQS permissions to staging developer IAM identity

P0NDER0SA:
## Summary

Follow-up to #2802. Live testing surfaced `AccessDenied` on `sqs:ListQueues`, and the team flagged additional gaps needed for local API/worker development against staging:

- `sqs:ListQueues` — required by `delete_sqs_queues.py`; has no resource-level permission support, so it must stay account-wide (`Resource: "*"`).
- `sqs:CreateQueue`, `sqs:PurgeQueue`, `sqs:DeleteQueue` — needed for `NOTIFICATION_QUEUE_PREFIX`-based dev queues and `run_celery_purge.sh`/`delete_sqs_queues.py`. These target queues that don't exist as Terraform resources, so they're scoped to the account/region ARN pattern (`arn:aws:sqs:ca-central-1:${account_id}:*`) rather than the fixed queue list.
- `sqs:ReceiveMessage`, `sqs:DeleteMessage`, `sqs:ChangeMessageVisibility` — required for Celery workers to poll/ack/extend visibility on the existing named app queues; added to the existing scoped `Sqs` statement.

## Validation

- `terraform fmt -check` passed.


ben851 (approved):
LGTM

